PR #5, opened 2017-01-26
Vobject objects can be passed to Event

Phyks:
Handle the case where the data passed to an Event is already a Vobject
object, to avoid back and forth conversion between iCal and vobject

This is to be able to pass a `vobject.VEvent` object directly to the
`add_event` method and do not rely on back and forth conversion between
`vobject` structures and ical strings, as proposed in
[#41](https://bitbucket.org/cyrilrbt/caldav/issues/41/vobject-dependency-situation).

@tobixen This PR should cover and replace https://bitbucket.org/cyrilrbt/caldav/pull-requests/22/. I also converted the `.hgignore` file to a `.gitignore` file and removed it. I am not 100% sure about the coding guidelines for this repo, whether it should be PEP8 or not. My editor is configured to remove trailing whitespaces, according to PEP8. I can revert it if you want. Putting some indications in a `CONTRIBUTING.md` might be useful :)


tobixen:
I'm not a big fan of PEP8, though it makes sense to adhere to coding standards :-)  I should probably have a look into why my editor leaves so many silly trailing whitespaces behind ... or write some commit-hook to strip it.

tobixen:
The biggest problem with piggybacking lots of whitespace-fixes or pep8-fixes into some commit or pull-request is that it makes it difficult to merge together different changesets later.  

Came to think, if we are to do semantic versioning the right way, this pull request should eventually go into 0.6, while bugfixes should go into 0.5.1 - that will be tricky due to the white space fixes.  ATOH, semver.org states that "anything may change at any time" with version numbers starting with 0, and this change seems relatively pretty harmless.  I guess it will slip into the 0.5.1-release, if there every will be one ... and we should probably release 1.0 soon ;-)

Phyks:
> The biggest problem with piggybacking lots of whitespace-fixes or pep8-fixes into some commit or pull-request is that it makes it difficult to merge together different changesets later.

Sure. I am not really a hg guru, so I'd rather leave it as is when backporting on Github. Next time, I'll split this in a separate commit.

Thanks for the merge!

PR #10, opened 2017-05-08
Reformat code according to PEP8

jammon:
This is largely a visual reformatting of the code according to PEP8. Most changes concern line length, trailing spaces, spaces around operators and the sorting of imports. There should be no functional changes.

Just one little thing: I replaced  `python_utilities.isPython3()` by `six.PY3`, which functionally equivalent. Probably I should have put this into a different pull request. Well.


jammon:
Great, thank you!

tobixen:
I just ran into one more argument against line wrapping: it makes grepping more difficult.  Ref #12, I wanted to locate all places in the code where headers are passed to the DAVClient.request method.  "grep -r request . | grep headers" didn't give me much.

jammon:
Sure, granted, that's an inconvenience. `grep -r --after-context=1 request . | grep --before-context=1 headers` (or shorter `grep -r -A 1 request . | grep -B 1 headers`) should do what you wanted, but at the cost of being more complicated.

Issue #11, opened 2017-05-29
Bug with URL quote/unquote


andrei-mart:
in caldav/objects.py:
DAVObject._handle_prop_response() unqotes key it returns:
line 174:            
       href = unquote(r.find('.//' + dav.Href.tag).text)
but in DAVObject.get_properties it is compared to URL path, which is quoted:  
lines 209-218: 
        path = self.url.path
        exchange_path = self.url.path + '/'

        if path in list(properties.keys()):
            rc = properties[path]
        elif exchange_path in list(properties.keys()):
            rc = properties[exchange_path]
        else:
            raise Exception("The CalDAV server you are using has "
                            "a problem with path handling.")
So, if base path contains quoted characters, they do not match and the server is blamed.
Following modification makes it working for me:
lines 209-210: 
        path = unquote(self.url.path)
        exchange_path = path + '/'

Also, a small optimisation in lines 212 and 214:
        'in properties' is equivalent of 'in list(properties.keys())', but more effective.

tobixen:
I'm sorry to say that I have no capacity for the foreseeable future to look into this (though I'm not able to foresee much of the future - in the very best case, perhaps I will have time in some few weeks).  Perhaps you could make a pull request?  I think it's important to include some test code that throws an error with the current code in the pull request.

I'm generally having a low threshold for accepting pull requests, though I will do more thoroughly testing before the next release.

andrei-mart:
I am not sure I will find time to put together proper fix (with test), but I will try.

JeyDee:
I can confirm that this solves the problem (thank you very much for this, andrei-mart).
Can you please accept the pull request, tobixen?

tobixen:
pull request has been accepted, but I still didn't get to release a new version

JeyDee:
OK, I see. Thanks :).

tobixen:
I believe this was fixed and that this issue can be closed

PR #13, opened 2017-06-02
Unquote URL path which is matched against property key, which is also…

andrei-mart:
… quoted

Minor optimization of search in a dictionary
ref https://github.com/python-caldav/caldav/issues/11

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -206,12 +206,12 @@ class DAVObject(object):
         rc = None
         response = self._query_properties(props, depth)
         properties = self._handle_prop_response(response, props)
-        path = self.url.path
-        exchange_path = self.url.path + '/'
+        path = unquote(self.url.path)
+        exchange_path = path + '/'
 
-        if path in list(properties.keys()):
+        if path in properties:
             rc = properties[path]
-        elif exchange_path in list(properties.keys()):
+        elif exchange_path in properties:
             rc = properties[exchange_path]
         else:
             raise Exception("The CalDAV server you are using has "


andrei-mart:
There is a problem to add a test.
The test would depend on server setup, and test servers are not a part of the test suite.
The environment I tested against uses email addresses as user names, and user name is also a part of base path, and '@' in the key returned from _handle_prop_response() was quoted. I got the error in very basic case: principal.calendars() failed. Current test suite has that call, and I expect it will fail if one of the test environments is set up as mentioned above. It may be enough to create user with name containing @ or # or & or < or > or any other character that should be quoted in URL.

tobixen:
Maybe it's possible to create a calendar with @ in the name?

andrei-mart:
I tried to add @ to IDs, tests pass anyway.
Should it be in the name? 
We want it to appear in the path when a property is returned. I am not familiar with caldav protocol and have no idea how to achieve that.

tobixen:
I'll merge this and look more into it when I get the time.

Achimh3011:
I just ran into this problem after a zimbra upgrade (user is identified by mail address). The PR solves this.

Everything fine, except: when will a new version be released on PyPI? I hate it when I need to reference package dependencies as git repositories...

tobixen:
I hoped to get time to look more into caldav previous week, but ... time is passing so quickly.  I got into some test failures, it seems like one of my caldav test servers is just dropping VTODO items rather than saving them.  Once this issue is fixed I can update the changelog and work on a new release.

Achimh3011:
Thx for the heads-up.

PR #15, opened 2017-06-14
Bugfix: Calendar.todos() now returns todos without a status

### LINKED ISSUE #14: Calendar.todos() ignores todos without status on NextCloud/OwnCloud and sabre/baikal
It seems like the Nextcloud/Owncloud tasks app does not set a status on new tasks. These tasks are not returned using `Calendar.todos(include_completed=False)`. If `include_completed` is set to True, they are returned as expected.

mayjs:
This PR fixes #14. The fix was quite simple but seems to do the trick.


tobixen:
I assume it has been tested.  It's considered best practice to also write some test code that would fail without the pull request, but anyway I'll merge it in now.

tobixen:
I'm reverting this out as for now, ref #14

Issue #3, opened 2017-01-24
Icloud not fully supported


tobixen:
# Updated summary

Updated 2020-11-23.

I've been running through the tests for iCloud now, and I have the following notes:

* Icloud uses calendar URLs like `https://p12-caldav.icloud.com/17112341234/calendars/12345@example.com/` - the numbers (even in the host name) differs from account to account.  There is no need to wonder too much about this, specify the caldav url to be `https://caldav.icloud.com/`, and the library can figure out the rest - code examples at https://github.com/python-caldav/caldav/issues/3#issuecomment-731650816
* The test code is (by default) creating and deleting the same calendar repeatedly.  I did observe some times that events stored on the calendar would not disappear if the calendar was deleted and then recreated.  I consider that to be a bug, but a bug that never should be encountered in ordinary use cases.  (could be that the bug is in the test code - did not investigate much).
* It seems that one unique calendar object with one unique uid cannot be on two calendars at the same time.  I'm not sure what the RFC says about this - I think it's a common use case scenario to have one and the same event on multiple calendars, and then it ought to have the same uid also.  (could be that the bug is in the test code - did not investigate much).
* I had to skip testing involving journal entries, freebusy requests, tasks and recurring events - did not investigate where it fails or if we can make any workarounds.  Does iCloud support those things?
* A test doing a propfind failed and has been skipped.  Not sure if it's something wrong with the iCloud or if it's something wrong with the test.
* The method `object_by_uid` (as well as `event_by_uid`) fails.  I vaguely remember someone telling me there was an issue with it on Google as well.  Should be investigated, perhaps the logic should be rewritten or some workaround should be found for iCloud (and Google?).
* I had to do some minor workaround in the code for fetching event lists - the calendar URL is sometimes passed in the response together with the events, causing the library to handle it as an extra event. I haven't bothered checking the RFC, but wrote up some lines to check for this and throw it away.

Except for those things ... all tests passes by now.

Please let me know (or update this issue) if there are any outstanding problems, if you know anything about the stuff I didn't get to work, if you want me to prioritize to look into those things, or ... whatever.

tobixen:
The quick TLDR; people have reported various issues when using this caldav library towards the icloud.  I don't seem to have a working test-account towards the icloud at the moment, hence I'm not working actively on research here.  People using icloud is encouraged to do research.

(actually, I vaguely remember someone provided me with a test account once upon a while - it may be that I've simply lost the credentials for it).

[...]

PR #17, opened 2017-08-09
Add iCloud example code

kyloe:
Added example directory with code for connecting to iCloud. COntains a class to wrap around caldav which  provides discovery service to retrieve principal and calendar-home-set urls for iCloud accounts.


tobixen:
Ref https://github.com/python-caldav/caldav/issues/3

PR #21, opened 2017-11-20
Added Apple ical namespace and properties

m42e:
Adding Apples ical namespace to allow retrieving of the calendar color and order.


tobixen:
I have mixed feelings on adding support for "propretary" extensions on the standard in the library, but ... well, if it's useful for Apple users and won't hurt other users, then ... why not.

PR #26, opened 2017-12-28
[objects] Unfold lines before checking for UID

### LINKED ISSUE #25: Long UIDs cause object_by_uid() to return a false negative
```
            if not "\nUID:%s\n" % uid in data:
                # TODO: optimistic assumption, uid line is not folded.  We
                # need to unfold the content to be 100% sure that we won't
                # filter away true positives here.
                continue
```

earlchew:
Fixes #25

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -719,10 +719,12 @@ class Calendar(DAVObject):
             # Ref Lucas Verney, we've actually done a substring search, if the
             # uid given in the query is short (i.e. just "0") we're likely to
             # get false positives back from the server.
-            if not "\nUID:%s\n" % uid in data:
-                # TODO: optimistic assumption, uid line is not folded.  We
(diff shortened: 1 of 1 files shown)


tobixen:
It would be nice to add some test code for this

earlchew:
@tobixen I started on a test, but found that I could not get a clean pass on master as a baseline:

```
  File "/home/ubuntu/local/caldav/caldav/objects.py", line 155, in _query
    raise error.exception_by_method[query_method](errmsg(ret))
caldav.lib.error.PropfindError: 503 Service Unavailable
...
root: DEBUG: ############## test setup
root: DEBUG: <LogRecord: caldav, 10, /home/ubuntu/local/caldav/caldav/davclient.py, 74, "url: http://baikal-test.caldav-servers.tobixen.no/cal.php/">
```

PR #28, opened 2018-04-06
Replaced vobject by icalendar.

tonioo:
I recently added a calendar plugin to the project I work on (Modoboa, https://github.com/modoboa/modoboa-radicale) and during my tests, I realized vobject was extremely slow to parse recurring rules. Since I'm not the first one to report this behaviour, I've replaced it by icalendar which offers better performance.

### DIFF
--- a/caldav/__init__.py
+++ b/caldav/__init__.py
@@ -3,10 +3,6 @@
 from .davclient import DAVClient
 from .objects import *
 
-# possibly a bug in the tBaxter fork of vobject, this one has to be
-# imported explicitly to make sure the attribute behaviour gets
-# correctly loaded:
-from vobject import icalendar
 import logging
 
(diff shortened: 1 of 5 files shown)


tonioo:
I understand your concerns but in my case (and I guess it was the same for others if you already received such a PR), it's just unusable with vobject... That might be a huge refactoring but what about a backend mechanism? Let people choose vobject and icalendar?

tobixen:
Letting people choose between icalendar and vobject somehow while also marking vobject as deprecated in the documentation.  Perhaps finding new names for the methods giving access to the icalendar objects.  Perhaps importing the vobject library only when needed (since it is becoming an optional dependency)

tobixen:
I'll close this pull request, as the master branch now supports icalendar objects

PR #30, opened 2018-06-14
Fix: directive for timezone

Alghost:
Fix directive to represent Timezone.

before fixing, It cause getting wrong events various timezones.


tobixen:
This commit is actually broken - according to the standards the timezone should be given in UTC-time, and the "Z" should be added to make it explicit that the timestamp is in UTC time zone.  I've reverted and tried to make a proper fix, ref #41.

PR #39, opened 2019-02-21
Extended Event class

flixman:
Caldav seems not do play around with the objects itself (maybe I missed something). Therefore, I have extended the Event class to

a) provide a method to know if the current event and another's time frames intersect.
b) generate a new event whose time frame is the union of the current event's time frame and that of the other input.

In case I am doing something completely off (so that this functionality is already provided) please let me know.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -914,7 +914,44 @@ class Event(CalendarObjectResource):
     """
     The `Event` object is used to represent an event (VEVENT).
     """
-    pass
+
+    # return wheter or not the other event intersects with the current one
+    def intersects(self, other):
+        if not isinstance(other, Event):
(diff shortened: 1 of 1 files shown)


tobixen:
Sorry for the slow feedback on this one.

I have two minor issues with this one, the first is that I feel it may be a bit out of scope for the caldav library, as this does not relate to caldav client/server communication but rather the iCalendar objects.

The other thing, there has for a long time been talks that we should move from the vobject to the icalendar module. Two pull requests have been written even, but both of them rejected as they don't preserve backward compatibility.

If you vouch to maintain this, as well as to rewrite to support the icalendar module if needed, I will merge it.

tobixen:
Closing this one because of inactivity and because I feel it's slightly out of the scope of the caldav library.

PR #49, opened 2019-03-11
Include tests and changelog in pypi sdist tarball

scarabeusiv:
The changelog can be useful for people trying to find out what changed and tests can be executed by the vendors to validate their python stack if it works with the caldav release.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,3 @@
 include COPYING.*
+include ChangeLog
+recursive-include tests *


tobixen:
Oh, I thought I had merged this one already.

tobixen:
Bad timing, just pushed through a 0.6.1-tag and then I saw this one ... so it will only be included in the next proper release.

Though ... if the default is to not include tests and ChangeLog, why should caldav deviate from the default?  Or am I missing something?  (I'm not much familiar with pypi, admittedly).

scarabeusiv:
It is not default, pypi won't upload anything that is not python modules, any other files including license you simply need to specify in this file. And having tests and changelog can be useful.

PR #53, opened 2019-04-06
Fix #52 - ensure correct behaviour if pytz is not used.

### LINKED ISSUE #52: Calendar.date_search ignores dates given
dates passed into `Calendar.date_search`are supposed to get their timezones fixed if pytz timezones are employed. However, the second return statement (taking care of non-pytz case) is incorrectly indented, making the functione return `None` and thus removing the date filter.

Achimh3011:
Fix indent.


tobixen:
Ideally we should fix some test code to catch this bug

PR #56, opened 2019-04-21
Necessary for iCloud..

ZiroNL:
Caldav is not working with iCloud without these changes. There are probably better solutions than mine.


tobixen:
I have a feeling more research ought to be done here - but anyway, those changes seems harmless, and if it makes things work for you on iCloud, then it's probably great :)

moozer:
I'm seeing the same using davmail to connect to an exchange server.

this PR solves it for me too.

PR #58, opened 2019-05-15
Fix #52 for real and add a test.

### LINKED ISSUE #52: Calendar.date_search ignores dates given

Achimh3011:
The code so far differed from its documentation (I think it did not work that well in all circumstances).

It also used a wrong way to apply a pytz timezone to an existing datetime (using `replace`) that ignores the time dependency of timezone definitions in pytz (and the olsen DB in general). 

In the PR I used the `datetime` own functionality to localize the `datetime` (supposed to be available in python 2.7 as well). pytz is only (this time correctly) used in the test.

I'm quite sure the code is right now (for what is documented as its purpose), but given we are talking timezones here, I suggest to believe me only after a thorough review and discussion.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -9,14 +9,19 @@ caldav server, notably principal, calendars and calendar events.
 import vobject
 import uuid
 import re
-import datetime
-import tzlocal
+from datetime import datetime, date
 from lxml import etree
 
 try:
+    # noinspection PyCompatibility
     from urllib.parse import unquote
 except ImportError:
     from urllib import unquote
+
+try:
+    from typing import Union, Optional
+except:
+    pass
 
 from caldav.lib import error, vcal
 from caldav.lib.url import URL
@@ -28,24 +33,21 @@ def errmsg(r):
     """Utility for formatting a response xml tree to an error string"""
     return "%s %s\n\n%s" % (r.status, r.reason, r.raw)
 
-def _fix_tz(dt):
+
+def _add_missing_tz(dt):
+    # type: (Optional[Union[date,datetime]]) -> Optional[Union[date,datetime]]
     """
     This method takes a datetime or a date and returns the same.
 
-    * if the dt object is a date, return it.
-    * if the dt object is a datetime with tzinfo, return it.
     * if the dt object is a datetime without tzinfo, assume it's in localtime,
       set it as such and return.
+    * otherwise (especially if it is a date or a datetime with tzinfo), return it unchanged.
 
     This method is used when searching the calendar with a date range;
     datetime objects will later be converted to UTC in elements.cdav
-
-    See also https://github.com/python-caldav/caldav/commit/7878d65fa4553cf91dde26334e5794c0852613b5#commitcomment-33417016 for more details
     """
-    if dt is None:
-        return None
-    if hasattr(dt, 'tzname') and dt.tzname() is None:
-            return dt.replace(tzinfo=tzlocal.get_localzone())
+    if hasattr(dt, 'tzinfo') and dt.tzinfo is None:
+        return dt.astimezone()
     return dt
 
 class DAVObject(object):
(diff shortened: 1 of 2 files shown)


tobixen:
This is brilliant ... I had no idea that .astimezone() could be used like this, and I had no idea that a timestamp could be tagged with the correct local timezone without introducing the dependency on pytz.

Investigating a bit, actually the _add_missing_tz-method may not to be needed at all.  Here is the code in caldav/elements/cdav.py:

```python
def _date_with_utc_time(ts):
    ## If tzinfo is given, convert to UTC
    if (hasattr(ts, 'tzname') and
        ts.tzname() is not None and
        ts.tzname() != 'UTC'):
        ts = ts.astimezone(pytz.utc)
    ## If not, assume the timestamp is already in UTC
    ## (that's very naïve - but in object.py, we explicitly add the
    ## local tzinfo if no tzinfo is given)
    return ts.strftime("%Y%m%dT%H%M%SZ")
```

It was my assumption that `ts.astimezone(pytz.utc)` would break, or at least that it would not do any conversion if ts was a naive timestamp.  Assumptions is the mother of all problems ... I should have done research on this earlier.  It seems that .astimezone(pytz.utc) converts from localtime to UTC time if the timestamp is naive ...

    >>> import pytz
    >>> datetime.datetime.now().astimezone(pytz.utc)
    datetime.datetime(2019, 5, 16, 7, 11, 41, 770649, tzinfo=<UTC>)
    >>> datetime.datetime.now().hour
    9
    >>> datetime.datetime.now().astimezone(pytz.utc).hour
    7
    >>> datetime.datetime.now().tzinfo
    >>> datetime.datetime.now().tzinfo
    >>> import pytz
    >>> import datetime
    >>> datetime.datetime.now().astimezone(pytz.utc)
    datetime.datetime(2019, 5, 16, 7, 19, 29, 802710, tzinfo=<UTC>)
    >>> datetime.datetime.now().astimezone(pytz.utc).hour
    7
    >>> datetime.datetime.now().hour
    9
    >>> datetime.datetime.now().tzinfo is None
    True
    >>> datetime.datetime.now().astimezone(pytz.utc).tzinfo is None
    False
    >>> 

Maybe a better way would be to drop the _fix_ts completely and simplify the _date_with_utc_time(ts) like this:

    def _date_with_utc_time(ts):
        if (hasattr(ts, 'tzname') and
            ts.tzname() != 'UTC'):
            ts = ts.astimezone(pytz.utc)
        return ts.strftime("%Y%m%dT%H%M%SZ")

That would introduce a compatibility issue if anyone would send timestamps without tzinfo directly to elements.cdav and not through the higher level methods. (old code would assume it to be UTC).  However, I don't think it's a big deal since it's relatively short time since 0.6.0 was released.

Next question: can we get rid of the dependency on pytz completely?  We've found that it's possible to find the time delta between UTC and local time without using pytz.

Achimh3011:
I'd think that using naive `datetime` is always a bit naive. A mention in the release notes spelling out how to avoid the problem for direct users of `elements.cdav` should be fine. I'm in favour of dropping the `_fix_ts`function. 

In your proposed `_date_with_utc_time` function, I'd suggest that you omit the check for `tzname` attribute and directly test for `isinstance(ts,datetime)` only - `astimezone(timezone.utc)` will take care of everything.

As you only need the local time zone and UTC, without regard to historical changes in timezones and without ever finding out the name of your local timezone, there is no further need for pytz and tzlocal. Code will (as far as I can think) work fine with clients using it, but also without it.

tobixen:
datetime.timezone.utc? That's nice, didn't see that one.  Brilliant, then we can remove the pytz-dependency.

I'm extremely busy now, but I will accept a pull request speedily.

Achimh3011:
PR superseded by #59.

PR #59, opened 2019-05-17
Remove fix tz

Achimh3011:
As discussed in #58, the function can be removed altogether.

### DIFF
--- a/caldav/elements/cdav.py
+++ b/caldav/elements/cdav.py
@@ -4,7 +4,6 @@ from datetime import datetime, timezone
 
 from caldav.lib.namespace import ns
 from .base import BaseElement, NamedBaseElement, ValuedBaseElement
-import pytz
 
 
 def _to_utc_date_string(ts):
--- a/caldav/elements/cdav.py
+++ b/caldav/elements/cdav.py
@@ -1,20 +1,19 @@
 #!/usr/bin/env python
 # -*- encoding: utf-8 -*-
+from datetime import datetime, timezone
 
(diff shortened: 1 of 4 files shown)


tobixen:
Brilliant.  I'll merge this now, and hopefully I will get time to test towards all the server implementations and release a new minor version within some few days.

tobixen:
One comment ... there is no such thing as SOMEWHERE_REMOTE, as the library may be used by people all over the world, I think SOME_TZ would be a better variable name :-)

Achimh3011:
One can always discuss about a name - I'm not sticking to it. But the remoteness of the place ([Fernando de Noronha)](https://en.wikipedia.org/wiki/Fernando_de_Noronha) is one of its key selling points as a special tourist location, so I'll hope nobody will take offense.

tobixen:
Ah ... cool :-)

tobixen:
This pull request is not compatible with python2.7 (nor python3.5).  I don't think it's necessary to support python3.5, but python2.7 is important.

* There is no datetime.timezone in python2.7.
* tz is a mandatory parameter to datetime.datetime.astimezone (you're using it without a given time zone in the test code)
* one gets an explicit exception when doing a .astimezone(..) on a naive timestamp in python3.5 (assume it's the same with python2.7)

Since setup.py is a python script, I suggest to make a check there, if it's not possible to do datetime.datetime.now().astimezone() we add pytz and tzlocal as requirements.  Same in the code, import pytz.utc and tzlocal if needed.

Issue #57, opened 2019-05-06
Automated test runs on commits and pull requests


tobixen:
Depends on #46, as we need clean tests that can be run without too many dependencies.

tobixen:
travis is now running unit tests, and on python3 it's doing functional tests towards xandikos.  (I believe it should be possible to set up more complex testing, let travis spin up other local test server instances, i.e. through docker.  Also, we could probably try to spin up radicale on python2.  Not sure if it's appropriate to close this issue just yet ...)

tobixen:
Now I can close this issue :-)

PR #62, opened 2019-05-19
Fix use of datetime.now.

jelmer:
Fix use of datetime.now.

The top of objects.py has:

 from datetime import datetime

so further on in the file, code should use datetime.now() rather than
datetime.datetime.now().


tobixen:
I wonder how this happened.  I guess I should prioritize #57

Issue #68, opened 2019-05-29
0.6.2-release


tobixen:
@cyrilrbt, I think it's time for another minor release.

Quite some embarrassing bugs have been found in 0.6.1, I hope they are properly ironed out now.  To reduce the risk of new bugs entering the master branch, by now unit tests and functional tests are run every time commits are pushed to github, or pull requests are submitted. 
 Still, the 0.6.2-release is admittedly under-tested - the public test servers I was running towards are currently down.  Anyway, I believe the risk of introducing new bugs is low compared to the benefit of ironing out known bugs.

Here are the release-notes:

* The timezone handling introduced in 0.6 was not very well tested, thanks to feedback from Achim Herwig this has been fixed.  In python 3.6 and higher, it's possible to convert between local time and utc without introducing the pytz/tzlocal dependencies.
* Quite some other bugs have been found and fixed.
* Some work has been done with the test framework:
  * Since the external public test servers have been notoriously unreliable, by default the tests will not be executed towards the external public test servers (and they are all down as of writing)
  * Xandikos is written in python, can be spawned up easily from python and has no other run-time dependencies than write access to a file system.  Spawning up a xandikos server instance from the test code was an easy thing to do, so by default now the tests are executed towards Xandikos.  However, Xandikos is python3-only, so this only applies when running tests from python3.
  * Unit tests are now run on every commit pushed to github, and on every github pull request.  This should stop the most embarrassing bugs from entering the master branch.

CAVEAT 1: some of the timezone handling has been moved from object.py to elements/cdav.py, and a minor incompatibility with 0.6.1 has been introduced; if you were using cdav/elements.py directly (and not using the higher-level methods in objects.py) and passing naive timestamps to it, they would be treated as UTC-time in 0.6.1 while they are assumed to be localtime in 0.6.2.  It seems very unlikely that this incompatibility issue would hit anyone.
CAVEAT 2: The 0.6.2-release is under-tested, all the servers/accounts used when testing 0.6.0 is currently down or unavailable.  (Please help!  Check tests/conf_private.py.EXAMPLE).

cyrilrbt:
I'm kind of busy at the moment, just moved long distance, but I'll check it out by the end of the weekend at the latest. Though at first glance it sounds like a 0.7.0 to me, even though we never explicitly said we were doing semver. I think we should probably make that explicit. Again, this is just after a quick read of this issue only, I'll get back to you very soon!

-------- Original Message --------

tobixen:
[...]

PR #76, opened 2019-09-24
Add long description to the package

sblondon:
The goal of this PR is to provide a description at https://pypi.org/project/caldav/ in order to replace the current text:
```
The author of this package has not provided a project description
```

I don't know which command is used to create the package but, according to `python setup.py sdist bdist_wheel`, it should work.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include COPYING.*
 include ChangeLog
+include README.md
 recursive-include tests *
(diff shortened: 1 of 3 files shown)


tobixen:
Perfect ... we can test this out by releasing a new version 0.6.2, ref #68.  Just need to get it tested towards a caldav server first.  Unfortunately I don't have the time to set up any test servers right now.

sblondon:
Thanks for the quick merges! :)

PR #103, opened 2020-06-12
Fix errors with full url response

mzyy94:
When full url is included in the response shown like below, some errors are apperred.

```xml
<d:multistatus xmlns:d="DAV:" xmlns:c="urn:ietf:params:xml:ns:caldav">
    <d:response>
        <d:href>http://cal.example.com/home/bernard/</d:href>
        <d:propstat>
            <d:prop>
                <c:calendar-home-set>
                    <d:href>http://cal.example.com/home/bernard/calendars/</d:href>
                </c:calendar-home-set>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
    </d:response>
</d:multistatus>
```
_from an example shown in https://tools.ietf.org/html/rfc4791#section-6.2.1_

I fixed them with:
  - avoid `quote` from a target of `url.join` if it's full url because 'https://\~' will be encoded to 'https%3A//\~' occurred with a change of https://github.com/python-caldav/caldav/commit/8f0c0e5d7e4cdf88b5c2e4260f114b405d8e4085
  - use full url as key to get value from properties when nothing matched with path only key

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -98,6 +98,10 @@ class DAVObject(object):
             resource_name = properties[path][dav.DisplayName.tag]
 
             if resource_type == type or type is None:
+                url = URL(path)
+                if url.hostname is None:
+                    # Quote when path is not a full URL
+                    path = quote(path)
                 # TODO: investigate the RFCs thoroughly - why does a "get
                 # members of this collection"-request also return the
                 # collection URL itself?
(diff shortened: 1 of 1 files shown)


mzyy94:
I tried to write tests with a response includes full url like `<d:href>http://cal.example.com/home/bernard/</d:href>` but I couldn't configure radicale for this test use..

tobixen:
I should try to write up some unit tests that mocks up a response with absolute URLs - and it's also nice if we can run the functional tests towards a caldav server that gives absolute URLs.  What server do you use?

tobixen:
I added a unit test, and I've also verified that the unit test breaks towards 0.7.0.

I would still like to know what kind of caldav server you are using.

Issue #94, opened 2020-05-17
object.id should always work


tobixen:
object.id is sometimes set and sometimes not.

Move it to object._id and make some getter method that will dig into the data to find the uid if self._id is not set.

tobixen:
See also #104.

tobixen:
This is actually a duplicate of #96

tobixen:
To clarify, `event.id` should always work and return the uuid for `CalendarObjectResource` objects sans `FreeBusy`.  I'm still not sure if it makes any sense to have an id field in other `DAVObjects`.

tobixen:
I've seen that `event.id` cannot always be trusted.  We need some unit tests and functional tests on this, covering all kind of edge-cases.

PR #104, opened 2020-06-15
Set `id` property on `Todo`s at object instantiation

jscn:
Consider this a straw-man to elicit some feedback. I've implemented setting the ID just on the `Todo` object for now, but this could obviously be pulled up into the parent class with the appropriate abstractions.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -811,7 +811,7 @@ class Calendar(DAVObject):
             # attempting a match.
             item_uid = re.search(r'\nUID:((.|\n[ \t])*)\n', item.data)
             if (not item_uid or
-                    re.sub(r'\n[ \t]', '', item_uid.group(1)) != uid):
+                    re.sub(r'\n[ \t]', '', item_uid.group(1).rstrip()) != uid):
                 continue
             return item
         raise error.NotFoundError("%s not found on server" % uid)
@@ -1115,6 +1115,13 @@ class Todo(CalendarObjectResource):
     """
     The `Todo` object is used to represent a todo item (VTODO).
     """
+
+    def __init__(self, *args, **kwargs):
+        """Set the object's ID."""
+        super().__init__(*args, **kwargs)
+        if hasattr(self.vobject_instance.vtodo, "uid"):
+            self.id = self.vobject_instance.vtodo.uid.value
+
     def complete(self, completion_timestamp=None):
         """
         Marks the task as completed.
--- a/tests/test_todo.py
+++ b/tests/test_todo.py
@@ -1,0 +1,60 @@
+from unittest import TestCase
+
+from caldav.objects import Todo
+
+
+# example from http://www.rfc-editor.org/rfc/rfc5545.txt
+VTODO = """BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Example Corp.//CalDAV Client//EN
+BEGIN:VTODO
+UID:20070313T123432Z-456553@example.com
+DTSTAMP:20070313T123432Z
+DUE;VALUE=DATE:20070501
+SUMMARY:Submit Quebec Income Tax Return for 2006
+CLASS:CONFIDENTIAL
+CATEGORIES:FAMILY,FINANCE
+STATUS:NEEDS-ACTION
+END:VTODO
+END:VCALENDAR"""
+
+
+## a todo without uid.  Should it be possible to store it at all?
+VTODO_NO_UID = """
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Example Corp.//CalDAV Client//EN
+BEGIN:VTODO
+DTSTAMP:19980101T130000Z
+DTSTART:19980415T133000Z
+DUE:19980516T045959Z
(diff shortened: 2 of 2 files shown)


caldav/objects.py
```diff
@@ -811,7 +811,7 @@ def object_by_uid(self, uid, comp_filter=None):
             # attempting a match.
             item_uid = re.search(r'\nUID:((.|\n[ \t])*)\n', item.data)
             if (not item_uid or
-                    re.sub(r'\n[ \t]', '', item_uid.group(1)) != uid):
+                    re.sub(r'\n[ \t]', '', item_uid.group(1).rstrip()) != uid):
```
jscn, line 814: I go this while debugging:
```
(Pdb) item_uid
<re.Match object; span=(334, 373), match='\nUID:19970901T130000Z-123405@host.com\r\n'>
(Pdb) item_uid.group(1)
'19970901T130000Z-123405@host.com\r'
```
It looks like there is a problem in the regex when parsing Windows-style line endings. The ID returned from `Todo.vobject_instance.vtodo` is `"19970901T130000Z-123405@host.com"`, but here the extra `\r` character means the IDs do not match. This change didn't seem to introduce any errors in the rest of the tests, but I don't have enough context to know if this will cause issues elsewhere (obviously it's a pretty ham-fisted fix, anyway). I would want to add some unit tests for this method before comitting a change here.

caldav/objects.py
```diff
@@ -1115,6 +1115,13 @@ class Todo(CalendarObjectResource):
     """
     The `Todo` object is used to represent a todo item (VTODO).
     """
+
+    def __init__(self, *args, **kwargs):
+        """Set the object's ID."""
+        super().__init__(*args, **kwargs)
+        if hasattr(self.vobject_instance.vtodo, "uid"):
```
jscn, line 1122: Presumably we'd want to handle `icalendar` data here, too?

More generally, I don't know if parsing the data in the `__init__` is the best way forward. Typically I'd implement separate factory functions for creating domain objects from raw data, rather than keeping that logic on the domain object itself.

tests/test_todo.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TodoTestCase(TestCase):
+
+    def test_it_sets_the_id(self):
+        todo = Todo(data=VTODO)
+
+        self.assertEqual(todo.id, "20070313T123432Z-456553@example.com")
+
+    def test_it_prefers_the_id_of_the_vtodo(self):
+        # Is this actually the expected behaviour?
```
jscn, line 47: This seems reasonable to me, but there may be some reason to prefer the `id` that's passed in to the constructor.

tobixen:
Oh, this one slipped through, sorry for the delay. I'm at vacation, will look more into it in some few days.

tobixen:
"some few days", hah.  Maybe next week :-)

caldav/objects.py
tobixen: good catch

caldav/objects.py
tobixen: I'm not used to factory functions, I should probably read up on the benefits of using such

tobixen: for sure we should handle icalendar also here yes, good catch.

tests/test_todo.py
tobixen: I think that if the caller sends one id as parameter to __init__() and another id implicitly in the oject passed, then the caller is doing something wrong, and perhaps an exception would be more appropriate?

tobixen:
So, to sum it up:

* As you've already hinted to in your own "review", icalendar data should be treated equal as vobject data when sent to the constructor.
* As you've already hinted to in your own "review" (and what strikes me just by reading the title of the pull request), this logic applies as much to events and journals as todos, so it should be implemented in the parent class.caldav/caldav/
* I'm not sure if it gives any point to give preferance to one or the other if someone calls a 

Other thoughts (could be raised as separate issues perhaps):

*  The logic does not apply to FreeBusy-responses, those are another beast entirely, as noted in the comments there is quite some of the stuff in the CalendarObjectResource is probably broken for FreeBusy-responses as it is, we probably need a CalendarResponseResource, probably as a parent to CalendarObjectResource - but fixing this is certainly outside the scope of this pull request - it's just needed to make sure nothing breaks when passing an icalendar or vtodo without an ID to the constructor.
* Factory methods vs __init__ ... that's something I should read up on, maybe it makes sense, but that would for sure be a separate issue or merge request.

caldav/objects.py
tobixen: There was also another issue report about breakages due to windows line endings, so it I've simply stripped out all instances of `\r` in `DAVClient.DAVResponse.__init__`, though since this is icalendar-specific, it should probably be put into another layer of the code.

tobixen:
I will close this.

The windows line endings have been dealt with in issue #112.

There is an outstanding issue #94 to fix the id-field properly.  It's also set in `_create`, and should be cleaned out from there before it's added in `__init__`.

Since we'll move from vobject to icalendar in version 1.0, I'd like to procrastinate this until 1.0.

PR #109, opened 2020-08-29
Enable tests of recurring events against Xandikos

jelmer:
Enable tests of recurring events against Xandikos.

### DIFF
--- a/tests/compatibility_issues.py
+++ b/tests/compatibility_issues.py
@@ -6,13 +6,6 @@
 ## in the documentation on everything.
 
 xandikos = [
-    ## Xandikos does not support recurring events as of 0.2.1/2020-05,
-    ## ref https://github.com/jelmer/xandikos/issues/8
-    "norecurring",
-
-    ## As of 2020-05-17 it even returns a 500 internal server error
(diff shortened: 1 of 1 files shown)


tobixen:
Hurray!

tobixen:
But maybe the test requirement list in setup.py needs to be updated to reflect the minimum version of xandikos to be used?

jelmer:
That's a good point. I'll release a new version this weekend and will send you a pull request to bump the minimum version.

tobixen:
What's the status on this?  I had to reintroduce the incompatibility flags as the tests breaks with 0.2.3 - and 0.2.4 and 0.2.5 breaks even more.

jelmer:
On Wed, Mar 10, 2021 at 02:25:08PM -0800, Tobias Brox wrote:
> What's the status on this?  I had to reintroduce the incompatibility flags as the tests breaks with 0.2.3 - and 0.2.4 and 0.2.5 breaks even more.
This is on my list of things to look at - will hopefully get to it
this week.

Jelmer

-- 
Jelmer Vernooĳ ***@***.***>
PGP Key: https://www.jelmer.uk/D729A457.asc

PR #110, opened 2020-08-29
Fix mocked method - davclient is now using requests

jelmer:
Fix mocked method - davclient is now using requests.Session.request rather than requests.request.


tobixen (commented):
I'm not at my laptop now, will merge now and verify it later.

tobixen:
That's actually the issue. I merged the session change request without verifying it. I should probably be a bit more careful. Or, better, try to figure out why automated testing doesn't work.

PR #111, opened 2020-09-06
Fixed a problem on certain caldav server (LINE Works).

mtorange:
This server returns empty data result (URL only) from date search.
So we have to re-query the data for every event after fetching urls. calendar_multiget() has been added for it.
And the exception on _calendar_comp_class_by_data() on empty data result has been fixed

### DIFF
--- a/caldav/elements/cdav.py
+++ b/caldav/elements/cdav.py
@@ -41,6 +41,11 @@ class FreeBusyQuery(BaseElement):
 class Mkcalendar(BaseElement):
     tag = ns("C", "mkcalendar")
 
+class CalendarMultiGet(BaseElement):
+    tag = ns("C", "calendar-multiget")
+
+
+
 
 # Filters
 class Filter(BaseElement):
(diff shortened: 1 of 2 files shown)


tobixen (commented):
Is it possible to create some test code for this?

mtorange:
> Is it possible to create some test code for this?

below is my working code.  
but I have no idea how can I create test code (we need specific CalDAV server for testing)


```
def get_events(url, user, password, calendar_name):
	start_dt = datetime.today() - timedelta(days=7 * 3)
	end_dt = datetime.today() + timedelta(days=7 * 12, hours=23, minutes=59, seconds=59)

	client = DAVClient(url, username=user, password=password)
	principal = client.principal()
	calendars = principal.calendars()

	for c in calendars:
		if calendar_name == c.name:
			events = c.date_search(start_dt, end_dt)
			if len(events) > 0 and events[0].data is None: # <-- server returns events with empty data
				events = c.calendar_multiget([ev.url for ev in events])
			return events
	return None

```

tobixen:
I'll merge this for now, but raise an issue that ...

* The multiget method should be exercised in test code, for servers that supports multiget
* Should add an extra parameter to date_search method to allow it to try to do a multiget if needed

PR #166, opened 2022-01-24
Fix digest auth with pre utf8 servers (version for master)

### LINKED ISSUE #164: Regression: Wrong credentials result in "AttributeError: 'DAVClient' object has no attribute 'password'"
This is on the "master" branch.
If the ```DAVClient```  constructor is called with wrong credentials, then a subsequent call to ```principal()``` results in an attribute error ```AttributeError: 'DAVClient' object has no attribute 'password'```  in line https://github.com/python-caldav/caldav/blob/cbca28592a4ccbf9f06601fdbbf2656e4afff17f/caldav/davclient.py#L536

Test case:
```python
import caldav

client =  caldav.DAVClient("https://example.com/some/path", username="some_user", password="some_secret")
user_principal = client.principal()
 ```
This was introduce by commit 164f88d8 and is caused by lines https://github.com/python-caldav/caldav/blob/cbca28592a4ccbf9f06601fdbbf2656e4afff17f/caldav/davclient.py#L548-L549 which should read instead:
```python
            self.username = None
            self.password = None
```

I found this while debugging another authorization problem for which a PR is coming soon and will fix this too, plus the other problem I was actually investigating.

### LINKED ISSUE #165: Digest authentication fails with some (really old) servers
Commit 13a4714 introduced proper handling for multi-lingual characters.  Unfortunately this breaks authentication with some really old pre-UTF8 servers, For example servers based on SabreDAV 3.1.x

See  PR #166 and #167 for a proposed solution.

BjoKaSH:
Some servers, like really old SabreDAV based servers, seem to choke on password if we pass a bytes sequence instead of a string to the "requests" libraries HTTPDigestAuth object.

This fix against master retries authentication in case of authentication failure with the same password as plain string.

Related: #167 (same fix, but for v0.8-stable)

Fixes #165 
Fixes #164

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -545,8 +545,8 @@ class DAVClient:
                 self.auth = requests.auth.HTTPDigestAuth(self.username, self.password)
             else:
                 raise NotImplementedError("Auth method %s not supported yet" % auth_type)
-            del self.username
-            del self.password
+            self.username = None
+            self.password = None
             return self.request(url, method, body, headers)
 
         # this is an error condition that should be raised to the application
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -545,6 +545,28 @@ class DAVClient:
                 self.auth = requests.auth.HTTPDigestAuth(self.username, self.password)
             else:
                 raise NotImplementedError("Auth method %s not supported yet" % auth_type)
+            return self.request(url, method, body, headers)
+
+        elif (r.status_code == 401 and
+            'WWW-Authenticate' in r.headers and
+            self.password and
+            self.username and
+            self.auth):
+
+            ## Some (ancient) servers don't like UTF-8 binary auth with Digest authentication.
+            ## An example are old SabreDAV based servers.  Not sure about UTF-8 and Basic Auth,
+            ## but likely the same.  so retry if password is a bytes sequence and not a string
+            ## (see commit 13a4714, which introduced this regression)
+
+            auth_type = r.headers['WWW-Authenticate']
+            auth_type = auth_type[0:auth_type.find(" ")]
+
+            if hasattr(self.password, 'decode'):
+                if auth_type == 'Basic':
+                    self.auth = requests.auth.HTTPBasicAuth(self.username, self.password.decode())
+                elif auth_type == 'Digest':
+                    self.auth = requests.auth.HTTPDigestAuth(self.username, self.password.decode())
+
             self.username = None
             self.password = None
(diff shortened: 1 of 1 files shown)


caldav/davclient.py
```diff
@@ -539,14 +539,38 @@ def request(self, url, method="GET", body="", headers={}):
             auth_type = r.headers['WWW-Authenticate']
             auth_type = auth_type[0:auth_type.find(" ")]
 
+            error.assert_(auth_type in ('Basic', 'Digest'))
```
tobixen, line 542: I think this is moot?  There is a `raise NotImplementedError` further below that handles unsupported auth types

caldav/davclient.py
```diff
@@ -539,14 +539,38 @@ def request(self, url, method="GET", body="", headers={}):
...
+
             if auth_type == 'Basic':
                 self.auth = requests.auth.HTTPBasicAuth(self.username, self.password)
             elif auth_type == 'Digest':
                 self.auth = requests.auth.HTTPDigestAuth(self.username, self.password)
             else:
                 raise NotImplementedError("Auth method %s not supported yet" % auth_type)
-            del self.username
```
tobixen, line 548: I thought it should be set to None?  I don't remember why I put in those lines right now, but there was a reason for it ... hm.

tobixen: Ah, nevermind - I see it's done further below

caldav/davclient.py
BjoKaSH: Hmm, I *think* this is probably a copy&past error left over from the massive merge conflict I got when attempting to rebase my v0.8 fix on top of master.  After which I did a reimplementation from scratch.

The line should be removed.  I'll refresh my PR accordingly

caldav/davclient.py
BjoKaSH: Removing these two lines altogether (and not yet setting ```self.password``` and ```self.username``` to ```None```) is necessary, since both values are needed for retry in the next iteration of the call in the next ```elif ...``` branch.
By the way, it took me a while to realize that you deal with authorization check and final authentication by recursively calling ```self.request()```, with implicit state tracking in ```self.auth``` and ```self.username```.  Elegant but a bit unintuitive.

tobixen: If it wasn't intuitive, then probably it wasn't elegant :-)  (but at least better than the approach in v0.8.2).

I'm quite annoyed spending so much time on this - I think this logic should have been in the requests library, as it is neither DAV-specific nor CalDAV-specific.

PR #167, opened 2022-01-24
Fix HTTP Auth failure with old (pre UTF-8) servers (version for v0.8 stable)

BjoKaSH:
Some servers, like really old SabreDAV based servers, seem to choke on
password if we pass a bytes sequence instead of a string to the
"requests" libraries HTTPDigestAuth object.

This fix against v0.8-stable retries authentication in case of authentication failure with
the same password as plain string.

Related: #166 (same fix, but for master branch)

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -518,6 +518,7 @@ class DAVClient:
 
         (combined_headers, proxies) = self._pre_request(url, body, headers)
 
+        auth_type = None
         if not self.auth:
             ## Try a test request w/o auth
             resp = self.session.request(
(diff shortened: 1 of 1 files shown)


tobixen:
Someone told me that they did have some authentication problems with the 0.8-version while master works - so I've decided to discontinue work on the 0.8-branch and release 0.9 soon, so this work is basically moot - but I'll merge it anyway :-)

BjoKaSH:
Thanks for merging, even when v0.8 is basically End of Life.

Note that I tried with both, master and v0.8, and both failed to authenticate against the server in question.  That's why I produced two patches, one for master and one for v0.8.  So #166 is still needed.

PR #175, opened 2022-03-23
add ctx manager support for davclient (PEP 343)

neonfighter28:
This PR adds support for PEP 343 

It makes sense to be able to use with-statements with the DAVClient object, as some might want to close the DAVClient regardless of whether there were any errors. One would have to use a try-finally statement, but PEP 343 recommends using a with-statement

Currently the user has to invoke `DAVClient.session.close()` to fully close the `requests.Session` object. This PR adds support for the following Syntax:

```python3
with DAVClient(url) as dav:
    ...
```
If any error occurs during the with block, the requests session will still be closed regardless

Another improvement should be a publicly available `DAVClient.close()` function which closes the `requests.Session()` adapter, otherwise one has to use `DAVClient.session.close()`

Kind regards
neonfighter28

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -337,6 +337,12 @@ class DAVClient:
 
         self._principal = None
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.session.close()
+
     def principal(self, *largs, **kwargs):
         """
         Convenience method, it gives a bit more object-oriented feel to


neonfighter28:
I will write an example in the documentation for this at a later point. I don't think test cases are needed for this? Let me know if you would like some nonetheless

tobixen:
Thank you, this will go into 0.9, which I hope to find time to release in the near future.

Test code would be good, eventually it would be nice to have 100% test coverage

neonfighter28:
There really isn't any test code to write. I drafted something, which is really unnecessary and I have no clue where it would go in the test file, as I'm not really familiar with unittest.

```python3
    def testContextManager(self):
        """
        ref https://github.com/python-caldav/caldav/pull/175
        """
        cal_url = "http://me:hunter2@calendar.example:80/"
        with DAVClient(url=cal_url) as client_ctx_mgr:
            ...

        client_regular = DAVClient(url=cal_url)

        assert_true(isinstance(client_ctx_mgr, DAVClient))
        assert_true(isinstance(client_regular, DAVClient))
```
 There isn't really much to test for this so maybe just check that it is an instance of the DAVClient class? If you have any input on this, please let me know and I will do my best to improve this

tobixen:
The test breaks without this patch, so it's good enough :-)

tobixen:
I've done a very small patch to the documentation.  The example code should also be reviewed and rewritten, but that will be for the 0.10-branch.  I'll split it out as a separate issue.

tobixen:
That would be nice.

What happens if the session isn't closed?  (because up until now, I never considered the need to bother with it) :-)

PR #176, opened 2022-04-05
Ignore case of `auth_type`

TheMcs000:
I am using mailcow behind a reverse proxy and mailcow returns the auth_type "basic" in lower case.

Therefore I propose that the match should be case insensitive

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -543,11 +543,11 @@ class DAVClient:
             self.username and
             not self.auth):
             auth_type = r.headers['WWW-Authenticate']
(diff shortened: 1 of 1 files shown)


tobixen:
Test code would be good, I do wonder what kind of web servers are using other casing than the normal, I do believe "Basic" and "Digest" is correct according to the relevant RFCs ... but whatever ... I can't see anything going wrong with this patch :-)

dubtec:
I am using Baikal on Nginx (since You're asking "what kind of web servers") and am facing the AuthorizationError due to this issue, too.

PR #193, opened 2022-06-29
Support for multiple authentication methods server headers

### LINKED ISSUE #192: Multiple authentication methods support
Currently, only the first authentication method supported by a server is tried:

https://github.com/python-caldav/caldav/blob/c19abb303dc915cd8471628ef0a4f034204d1b5b/caldav/davclient.py#L549

However, sometimes servers support multiple authentication method, e.g. `Bearer` and `Basic`, but caldav will not attempt to connect if the unsupported method appears first in the `WWW-authenticate` header. For instance, this is the header provided by my server:

```
Bearer realm="Foobar", Basic realm="Foobar", charset="UTF-8", Basic realm="Foobar", charset="UTF-8", Basic realm="Foobar", charset="UTF-8"
```

Instead of just looking for the first authentication method in the header, I suggest to iterate over all the methods in the header until one is found that does not raise a `NotImplementedError`.

What do you think?

azmeuk:
Fixes #192 

Allow the library to work with servers supporting multiple authentication methods, for instance when the `WWW-Authenticate` header returned is:

```
Bearer realm="Foobar", Basic realm="Foobar", charset="UTF-8", Basic realm="Foobar", charset="UTF-8", Basic realm="Foobar", charset="UTF-8"
```

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -498,6 +498,12 @@ class DAVClient:
     def options(self, url):
         return self.request(url, "OPTIONS")
 
+    def extract_auth_types(self, header):
+        auth_types = header.lower().split(",")
+        auth_types = map(lambda auth_type: auth_type.strip(), auth_types)
+        auth_types = map(lambda auth_type: auth_type[:auth_type.find(" ")], auth_types)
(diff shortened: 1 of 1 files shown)


tobixen:
Thanks

PR #194, opened 2022-06-29
Tox newer environments

azmeuk:
This patch fixes the tox unit test command, and adds tests for python 3.9 and python 3.10


tobixen:
Super!  I'm still using nosetest myself, but should seriously consider using tox instead.

PR #195, opened 2022-06-29
CI with Github Actions

azmeuk:
This patch bring GHA instead of travis, for a better CI integration with github.


tobixen:
Thanks so much.  Automatic testing from github is high on the wishlist, I think I have an open issue on it as well.

Issue #196, opened 2022-06-29
Use pytest for unit tests


azmeuk:
Hey o/
You said you are using nosetests to run unit tests, but nowadays pytest receive a lot more attention than nosetests, while being fully compatible.
What do you think of running tox and GHA tests with pytest?
If you are OK I can provide a patch for this.

tobixen:
Absolutely.  I did read somewhere that nosetests was deprecated.

tobixen:
I think this issue can be closed now?

azmeuk:
Yes absolutely :+1:

PR #197, opened 2022-06-29
use pytest instead of nosetests

azmeuk:
This patch:
- installs pytest and pytest-cov to run the tests
- configure tox so it runs with pytest
- replace nose style assertions `assert_equal`, `assert_not_equal` and `assert_raises` with their pytest equivalents `assert` and `with pytest.raises`
- fixes a couple of warnings raised by pytest about regex string missing `r` flag

Before nose can be totally removed, it is needed to replace the `SkipTest` exception with the pytest `@pytest.mark.skip` or `@pytest.mark.skipif` decorator equivalent.
https://github.com/python-caldav/caldav/blob/6b791de7ccae48901ada8318158ae824f8f36982/tests/test_caldav.py#L25
Though it is not as simple as I expected so I havn't implement this in this patch.

Fixes partially #196

### DIFF
--- a/caldav/lib/vcal.py
+++ b/caldav/lib/vcal.py
@@ -50,7 +50,7 @@ def fix(event):
     """
     ## TODO: add ^ before COMPLETED and CREATED?
     ## 1) Add a random time if completed is given as date
-    fixed = re.sub('COMPLETED:(\d+)\s', 'COMPLETED:\g<1>T120000Z',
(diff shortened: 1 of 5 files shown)


tobixen:
Thanks a lot.

I wonder if the `xandikos<0.2.4` still makes sense.  As far as I remember, I was spending quite some efforts mending the test framework to support newer versions of xandikos.  For all I know, support for elder xandikos may even be broken now.  (I don't have time looking into it right now)

tobixen:
Four checks passed.  I guess that means the old xandikos version still works :-)

PR #199, opened 2022-06-30
Automatic code linting with black and pre-commit

azmeuk:
This one is a bit more opinionated. If you don't like it, we can close the PR :)

It adds [pre-commit](https://github.com/pre-commit/pre-commit) configuration, that reorder python imports alphabetically, and applies [black](https://github.com/psf/black) to format the code. Black is a kind of formatting standard nowadays in python, and it is optimized to reduce the diff length when patches are submitted, and the conflict chances when merging.

The style can be checked and fixed with `tox -e style`, and GHA also checks it.

What do you think?


caldav/__init__.py
```diff
@@ -1,18 +1,20 @@
...
 
 from .davclient import DAVClient
-## TODO: I don't think I like the practice of "import *" here.  Should be revisited prior to launching 1.0 at least.
 from .objects import *
 
+## TODO: I don't think I like the practice of "import *" here.  Should be revisited prior to launching 1.0 at least.
 # possibly a bug in the tBaxter fork of vobject, this one has to be
 # imported explicitly to make sure the attribute behaviour gets
 # correctly loaded:
-from vobject import icalendar
```
tobixen, line 10: And here too ... :-)

Perhaps I should clean up the comments.

tests/compatibility_issues.py
```diff
@@ -12,269 +11,182 @@
...
-
-    'object_by_uid_is_broken':
-        """calendar.object_by_uid(uid) does not work""",
-
-    'no_mkcalendar':
-        """mkcalendar is not supported""",
-
-    'no_overwrite':
-        """events cannot be edited""",
+    "no_expand": """Server may throw errors when asked to do a expanded date search""",
```
tobixen, line 14: I'm not too happy with this formatting

azmeuk: It is possible to tell black to not format some parts of the code with `# fmt: off` and `# fmt: on`

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

tests/compatibility_issues.py
```diff
@@ -12,269 +11,182 @@
...
+    """and/or deleting a calendar may not have immediate effect""",
+    "object_by_uid_is_broken": """calendar.object_by_uid(uid) does not work""",
+    "no_mkcalendar": """mkcalendar is not supported""",
+    "no_overwrite": """events cannot be edited""",
 }
 
 xandikos = [
     ## TEMP TEMP TEMP - TODO - should be investigated
     ## (perhaps my xandikos version is too old?)
-    "no_expand", "no_recurring",
```
tobixen, line 149: Another comment context that gets broken by "correct" formatting.  The comment above applies to both no_expand and no_recurring, which are intentionally placed on the same line to make it easy to remove them at a later stage.

tests/compatibility_issues.py
```diff
@@ -12,269 +11,182 @@
...
     "no_freebusy_rfc4791",
-
     ## Expanding recurrent events is not yet supported
     ## ref https://github.com/Kozea/Radicale/issues/662
     "no_recurring_expandation",
-
-    'no_scheduling',
-
+    "no_scheduling",
     ## extra features not specified in RFC5545
```
tobixen, line 94: Here also,  the extra line breaks was intentionally put there to group together things.

tobixen:
It looked fine at the first glance, but studying the details I do see some issues ... well, I should probably spend some time to look into the style changes and either clean up comments or explicitly disabling auto-formatting for the parts of the code where auto-formatting breaks the intended readability of the code :-)

PR #204, opened 2022-08-02
Allow requests connection timeout to be set

schwma:
I run `caldav` requests every 15 minutes in an endless loop and eventually my program will freeze indefinitely. This is because by default the `requests` library uses a `timeout` value of `None` which equates to an infinite timeout value. This PR allows developers to instantiate `DAVClient` with an optional `timeout` value that is passed on to the `request` method call.

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -314,6 +314,7 @@ class DAVClient:
         username=None,
         password=None,
         auth=None,
+        timeout=None,
         ssl_verify_cert=True,
         ssl_cert=None,
     ):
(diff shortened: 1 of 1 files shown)


tobixen:
This (the request library having default infinite timeout) has bitten me in some other settings as well.  I think that infinite timeout is not a good idea, particularly for web requests.  Maybe the default ought to be something sufficiently high (i.e. 600 seconds) rather than None?

schwma:
I agree, a default timeout value would be useful and is what I, as a developer using the `caldav` library, would expect. It seems as though there even is a [discussion](https://github.com/psf/requests/issues/3070) about eventually adding a default timeout value to `requests` itself.

tobixen:
Good - though the details are in the blue, it seems to me that there is some kind of consensus that at some point in the future, the requests library will have some kind of default timeout set.  Though, in that case the default should not be `None`, the default should be to not pass the parameter (and let the requests libary set the default).

PR #208, opened 2022-08-17
make URL picklable

tubaman:
The pickle module calls `getattr`[1] when unpickling a URL.  Make it so
that `__getattr__` doesn't assume that `self.url_parsed` exists so that
unpickle will work.

[1] https://docs.python.org/3/library/pickle.html#pickle-inst

### DIFF
--- a/caldav/lib/url.py
+++ b/caldav/lib/url.py
@@ -100,6 +100,8 @@ class URL:
     # To deal with all kind of methods/properties in the ParseResult
     # class
     def __getattr__(self, attr):
+        if 'url_parsed' not in vars(self):
+            raise AttributeError
         if self.url_parsed is None:
             self.url_parsed = urlparse(self.url_raw)
         if hasattr(self.url_parsed, attr):
--- a/tests/test_caldav_unit.py
+++ b/tests/test_caldav_unit.py
@@ -8,6 +8,7 @@ to emulate server communication.
 
 """
 from datetime import datetime
+import pickle
 
 import caldav
 import icalendar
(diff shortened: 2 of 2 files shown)


tubaman:
Here's an example of a similar fix: https://stackoverflow.com/a/49380669/226697

tobixen:
I wonder if there exists any other library doing the same, this URL class doesn't have much to do with the caldav protocol.  My reason for making (IIRC) was that some places in the code a parsed URL was expected, while other places a string was expected ... it was quite a mess, and I wanted some consistency.

tubaman:
Maybe just use the [5-value namedtuple returned from urlsplit](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit)?

tobixen:
That's how it was - some methods expecting that nametuple and others expecting a string ... and when reading the comments on the URL class, some methods also expected a relative URL while others expected an absolute URL.

From a usability point of view, I think it should always be possible to pass a string for "public" methods, it would be pretty bad to require the end user to always parse the url through urlsplit before sending it to the library.  I also tend to believe that backward-compatibility is to be prioritized - hence, it should always work uot to send a string to methods that formerly expected a string, and it should always be possible to send a namedtuple to methods that formerly expected that.  And then there are some other problems with URL-handling that has been passed ... like, stripping username and password from the URL, comparing two URLs, etc.  In my opinion a proper URL class belongs to the python distribution.  I find it hard to believe that I'm not just reinventing the wheel here.  If nothing like this already exists, it could be an idea to split out the URL class and package it as an independent library.

Issue #128, opened 2020-12-29
DAVObject.name should probably go away


tobixen:
`DAVObject.__init__` takes a named parameter name, which (according to the docstring) should be the displayname.

As far as I know (I haven't done research on it), this is relatively moot - the name given is AFAIK used only when creating a new calendar (?).  Also, the "display name" is not a universal property for all DAV objects - hence it should be removed from there.

I recently added DAVObject.props recently which can be used for storing a display name (but then .save() and .load() should ensure to save/update such properties).

This is a refactoring thing that should be done prior to the 1.0 release; the API will change a bit as only calendar objects will accept a name property on initialization (and a NotFoundError should be raised if we're unable to load a calendar with the given name).

tobixen:
Principal also has a DisplayName.  Perhaps let name be a property that mirrors the DisplayName.  We already have a getter `get_display_name`.  We have things cached in `self.props`.

tobixen:
Some work was apparently done during the last weeks moving this one from DAVObject to Calendar, and the CHANGELOG states that this is fixed ... but thinking harder about this, perhaps not?

`DAV:displayname` is not a calendar-specific thing, it's a defined in RFC 4918 and may apply to any DAV object.

I think this may make sense:

* Keep the get_display_name() on the DAVObject layer.
* No name attribute anywhere, neither in calendar nor DAVObject.
* In internal methods where `object.name` is set, we should rather populate it in the `self.props` cache
* We should change get_properties so that it also can be used for *only* probing the cache (and returning None if nothing is in the cache)
* DAVObject.name should be a read-only deprecated property that will call on self.get_properties and return None.
* Usage of DAVObject.name is not core functionality, I'm unaware if anyone uses it, so we can slap a deprecation warning on this already in caldav 3.0.

PR #212, opened 2022-09-07
Added get_name convenience method for Calendar

mc-borscht:
Requested get_name convenience method created for Calendar object.

Returns name object attribute.

Docstring indicates that a None object is returned if a name is not set - maybe this is beyond the scope of this method's knowledge though?

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -1272,7 +1272,7 @@ class Calendar(DAVObject):
         """
         Get calendar display name
         """
-        return calendar.get_property(dav.DisplayName())
+        return self.get_property(dav.DisplayName())
 
 
 class ScheduleMailbox(Calendar):
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -1267,13 +1267,12 @@ class Calendar(DAVObject):
         root = cdav.CalendarQuery() + [prop, filter]
 
         return self.search(root, comp_class=Journal)
-    
+
     def get_display_name(self):
         """
         Get calendar display name
         """
         return calendar.get_property(dav.DisplayName())
-
 
 
 class ScheduleMailbox(Calendar):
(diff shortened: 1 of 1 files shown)


tobixen:
One would expect `Calendar.get_name()` to yield the "display name" if such a thing is set at the server side.  The `self.name` attribute is some artifact which is only set if given by the client.  Ref #128 `self.name` is going to be removed in a future release.

I don't mind having such a convenience method, but it should do the right thing - and that is to do a call on the `get_property` method.

(This comment supercedes my previous comment, so I will delete it)

mc-borscht:
Agreed, makes sense.

1. I've pulled the original method and replaced with a new method called **get_display_name** which follows the desired behavior.
2. Replaced all calls to get_property(dav.DisplayName()) with calls to the get_display_name method

Let me know if anything else required or if you want me to change the method name back to **get_name**

caldav/objects.py
```diff
@@ -398,7 +398,7 @@ def calendar(self, name=None, cal_id=None):
         """
         if name and not cal_id:
             for calendar in self.calendars():
-                display_name = calendar.get_property(dav.DisplayName())
+                display_name = self.get_display_name()
```
tobixen, line 401: `calendar.get_display_name()` obviously :-)

tobixen:
Don't worry about style test not passing.  I believe it can be auto-resolved by running `tox -e style`.  I will eventually do it at some point, but of course for annotation purposes it's better that you do it.

mc-borscht:
Minor formatting changes were carried out by `tox -e style` command as suggested - it passes when I re-run the command.

tobixen:
There is still a bug on line 401

mc-borscht:
Fixed the following issues and passed style test

* Bug in get_display_name method - incorrectly call of get_property instance method
* get_display_name called by each calendar variable in CalendarSet calendar method - as pointed out by your comment

Hope this closes things - sorry about that!

PR #228, opened 2022-11-03
Drop python2 support

danigm:
Python 2 went EOL on 1/1/2020 and we receive no more support from the Python Core Team after Python 2.7.18 even for known security issues.

This removes the six package from dependencies.

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -2,6 +2,7 @@
 # -*- encoding: utf-8 -*-
 import logging
 import re
+from urllib.parse import unquote
 
 import requests
 from caldav.elements import cdav
@@ -20,8 +21,6 @@ from caldav.objects import ScheduleInbox
 from caldav.objects import ScheduleOutbox
 from caldav.requests import HTTPBearerAuth
 from lxml import etree
-
-from urllib.parse import unquote
 
 
 class DAVResponse:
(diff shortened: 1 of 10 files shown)


tobixen:
Currently the caldav library is in a bit of a limbo status when it comes to python2, at one hand there is code in place to support python2, at the other hand I haven't done any testing towards python2 for quite some releases now - so the python2-support there is is likely to be broken.  From one perspective "no support" may be better than "probably broken support"  My official stance on this is that if I get bug reports that it breaks on python2, I will still fix it.

While python2 isn't maintained by the Python core team, it still exists as a package for Ubuntu 20.04 which is EOL'ed in 2030 - and 2.7 is the default python version in CentOS7, which is EOL'ed in 2024.  Theoretically, Canonical and RedHat are supposed to take the responsibility of patching up python if some security issue should appear.

Everything considered, this pull request will be put on hold for some time - perhaps not all the way until 2030, but at least until 2024 :-)

tobixen:
Came to think, python2-support must have been efficiently broken for some time already due to dependencies on other libraries.

I will make a release (1.0) within some few days, as soon as I'm done with that I will roll out this pull request - unless there will be a queue of hard-code dinosaur python2-users  protesting.

PR #229, opened 2022-11-03
Remove nose dependency

danigm:
Nose was only used to raise SkipTest that can be done with pytest.skip and that old dependency can be removed.


tobixen:
Nice one.  I noticed myself pytest.skip being used in another project, so I would probably have been fixing this myself today or tomorrow ... but now I don't have to bother :-)

PR #236, opened 2022-11-29
date_search  method is replaced

yeshwantthota:


examples/basic_usage_examples.py
```diff
@@ -72,16 +72,16 @@
 ## (this may fail if the server doesn't support expand)
 print("Here is some icalendar data:")
 try:
-    events_fetched = my_new_calendar.date_search(
-        start=datetime(2021, 5, 16), end=datetime(2024, 1, 1), expand=True
+    events_fetched = my_new_calendar.search(
+        start=datetime(2021, 5, 16), end=datetime(2024, 1, 1),event= True, expand=True
```
tobixen, line 76: One space too much ... if you run `tox -e style`, it will autocorrect.  (I could do it, but better you do it to get it into the same commit in the master branch and with the correct annotation)

tests/test_caldav.py
```diff
@@ -1457,10 +1457,11 @@ def testTodoDatesearch(self):
         # t6 has dtstart and due set prior to the search window, but is yearly recurring.
         # What will a date search yield?
         noexpand = self.check_compatibility_flag("no_expand")
-        todos1 = c.date_search(
```
tobixen, line 1460: This code does not need to be touched.  If you look right below, the new `search`-method is already exercised in this test code.

tobixen: I believe this applies to all code in the test directory.  I went through there myself to make sure the `search`-method would do the same as `date_search`.

yeshwantthota: so in the tests directory, I should not make any changes.

tobixen: Correct.  The date_search should be removed eventually, but only after the 1.0-release.  Still need the tests in place in case of retention issues.

tobixen:
One missing new-line.  `tox -e style` should autocorrect that.  And I should try and see if I can run the example test code.  Except for that it's good to go.

yeshwantthota:
> One missing new-line. `tox -e style` should autocorrect that. And I should try and see if I can run the example test code. Except for that it's good to go.



> One missing new-line. `tox -e style` should autocorrect that. And I should try and see if I can run the example test code. Except for that it's good to go.

I have added a new line

tobixen:
Something broke when I was trying to run the `basic_examples.py`.  I will check it up a bit later.

tobixen:
Yes, forgot to say.  The `compfilter` attribute is replaced by `event=True` or `todo=True` (and I should probably supplement it with `journal=True as well).  If you add this (I could of course do it, but better you do it so we can have a nice commit log) we should be good:

```
--- a/examples/basic_usage_examples.py
+++ b/examples/basic_usage_examples.py
@@ -157,12 +157,11 @@ assert "FREQ=YEARLY" in todos[0].data
 print("Here is some more icalendar data:")
 print(todos[0].data)

-## date_search also works on task lists, but one has to be explicit to get them
+## date_search also works on task lists
 todos_found = my_new_tasklist.search(
     start=datetime(2021, 1, 1),
     end=datetime(2024, 1, 1),
-    compfilter="VTODO",
-    event=True,
+    todo=True,
     expand=True,
 )
 if not todos_found:
```
```

Issue #201, opened 2022-07-07
Skipped event with missing 'vevent' property (Home Assistant)


dreimer1986:
As there is no real reason to be found in the HA CalDAV module, I try to pass things over to your issues, too.

First, the open report: https://github.com/home-assistant/core/issues/74196
It includes Logs of caldav Python module on DEBUG, so maybe you have an idea?

The recent beta updated the dependencies to 0.9.1 and after that things started to change a bit. Before that the module was unable to connect to my work CalDAV server at all (Error 403), now it does progress further, but every parsing of the server causes thousands of "Skipped event with missing 'vevent' property" errors in Home Assistant. These are not due to really missing anything, but due to thousands of error 404 returned and thus empty entries being parsed. At least that is my interpretation of the logs. Sure if that is the case, that's a error that should be caught and handled, but the bigger question is, why does this happen? I am able to load some entries by hand with the entries in the log that did not error out. The server itself is working fine on Mac and a few Android apps for CalDAV, but I am very sure that it's everything, but RFC conformant. So it's not really a big flaw or something as the ones who should start to fix things are the creators of our CMS, but as I said, the apps I tested work fine with it.

If you need further information, I happily provide these. A test account would be quite impossible as it's our work calendar. We could meet with a Teamviewer session though if you need some more information I can not provide otherwise.

tobixen:
I was doing a quick grep for the word "Skipped" in the caldav library and found nothing.  Perhaps the error is coming from the vobject library.  I had a hypothesis that the 404 caused empty strings to be sent to the vobject library, but apparently not ... that seems to cause an exception ...

```python
>>> vobject.readOne("")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/vobject/base.py", line 1155, in readOne
    return next(readComponents(stream, validate, transform, ignoreUnreadable,
StopIteration
```

I'm not sure when I will have time to look more into this, quite busy days and my capacity today is a bit reduced - apparently I've caught some virus.

dreimer1986:
Take all time you need. I am in absolutely no hurry here. Just wanted to pass on the new behavior.

tobixen:
The error message itself comes from HomeAssistant.  I'll close this one - I think further research should be done in the HomeAssistant issue report, and then we can rather reopen this issue if the problem is in the caldav library.

tobixen:
[...]

PR #247, opened 2023-01-14
Workarounds for broken caldav servers

tobixen:
* Zimbra may apparently create events with both DTSTART, DTEND and DURATION, which is a breach of the standard.  A simple workaround has been implemented, just throwing away redundant information.  No check done on weather the DURATION matches DTEND-DTSTART (it did in the ical data I got from Zimbra), so this one may potentially throw away information.  (I just wrote a message to the organizer asking if the original ical data was actually produced internally Zimbra or by some other client).
* I've been fighting a bit with some proprietary system in #201 that have some partial and slightly non-compliant support for caldav - and made some partial workaround for loading content from it.
* One extra unrelated commit enhancing the tests code has been piggy-backed on this pull request.

### DIFF
--- a/caldav/lib/vcal.py
+++ b/caldav/lib/vcal.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- encoding: utf-8 -*-
 import datetime
+import logging
 import re
 import uuid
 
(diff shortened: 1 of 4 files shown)


tobixen:
I shouldn't do such things in the late evenings ... commit 22011e49d95b8db59f385af49149e8a5499e582a and 5ad53ce28efb49df0f344f4b297e5e0d142c9a0a should have been included in this pull request :-)

PR #284, opened 2023-02-18
Polishing before release

tobixen:
added some more entries to the list of servers that does not support search queries containing is-not-defined


tobixen:
Updated the changelog

Issue #285, opened 2023-02-21
Support passing custom HTTP headers


JasonSanDiego:
I'm using Nextcloud and want to retrieve calendar (read only) subscriptions along with the normal read/write calendars. Nextcloud supports two ways of doing this. The easier of the two is to pass the custom HTTP header: X-NC-CalDAV-Webcal-Caching: On

[More details](https://help.nextcloud.com/t/propagation-of-a-subscription-calendar/82539/1)

Are folks open to adding this functionality by modfying the DAVclient initialization function to have an optional header parameter?

Example use:
`client = caldav.DAVClient(..., headers={"X-NC-CalDAV-Webcal-Caching": "On"})`

If so, I can file a pull request with this small change.

tobixen:
Pull request is appreciated.

tobixen:
The feature will be released in 1.2 shortly.  However, the issue should not be closed, we're missing documentation and test code.

tobixen:
This was fixed in #306 indeed.  Thanks.  Will be released in 1.3 today, I hope.

PR #288, opened 2023-02-23
Allow passing custom HTTP headers to DAVClient

JasonSanDiego:
See #285


tobixen:
The style test can be easily fixed by running "tox -e style"

tobixen:
Oh, nevermind that, the style breakage is mine (the github merge queue system seems to merge things even if the style check breaks)

Issue #289, opened 2023-02-25
extract_auth_types - shortening strings


jdrozdnovak:
Hello\
This lambda function https://github.com/python-caldav/caldav/blob/master/caldav/davclient.py#L553 not works ideally when having simple headers like `www-authenticate: Basic`\
I don't know enough about other caldav environments to do a change myself. I have only 2 available for test.

what the `extract_auth_types` returns in my case is `['basi']` which fails the future `"basic" in auth_type` condition

Thanks for reviewing.

tobixen:
Weird - I would have expected some of my test servers to fail, and people from the homeassistant community to be whining a lot if basic auth was broken.

This code section seems to be missing unit tests, so I'll try to write up some and possibly see if I can reproduce the problem.

tobixen:
I can reproduce it.

tobixen:
This bug has been there since 1.0 actually, it wasn't introduced in 1.1.  I suppose almost all servers with basic auth have a realm in the header.

tobixen:
@jdrozdnovak ...

It would be nice if you could test and see if the master branch solves the problem for you - or, if it's possible to make me a testing account on your caldav server allowing me to test it myself it would be great.

Since this is a considered to be a critical bugfix, I will roll out a new minor-release soon - I should just run the test code first towards a handful of servers.  Anyway, I'm pretty sure the fix does not break anything.

jdrozdnovak:
I can tomorrow test it with master and let you know.
I am using monicahq, where it fails. 

docker run --name some-monica -d -p 8080:80 monica

tobixen:
Ok, thanks

jdrozdnovak:
I tested it with the code in master branch and that works fine.

tobixen:
I'm running tests and will release 1.0.4 during the day probably

jdrozdnovak:
thank you very much! Have great sunday!

PR #290, opened 2023-02-25
Fix for issue #289 - basic auth broken towards some servers

tobixen:


tobixen:
Background:

A bug was introduced in version 1.0, via https://github.com/python-caldav/caldav/pull/260 - the code would only work if there was a space in the `WWW-Authenticate` header.  This works for most servers as they will challenge for credentials using a header like `WWW-Authenticate: Basic realm="My CalDAV server"` - however, `WWW-Authenticate: Basic` is fully allowed by RFC2617

PR #299, opened 2023-03-08
fix log message : argument was bytes not String

aaujon:
Fix logger, error message was :

```
  File "/home/arnaud/.local/lib/python3.10/site-packages/caldav/davclient.py", line 77, in __init__
    "Expected some valid XML from the server, but got this: \n"
TypeError: can only concatenate str (not "bytes") to str
```

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -70,7 +70,7 @@ class DAVResponse:
                 except:
                     logging.critical(
                         "Expected some valid XML from the server, but got this: \n"
-                        + self._raw,
+                        + str(self._raw),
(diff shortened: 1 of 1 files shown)


tobixen:
While you're at it, a bit further up there is this one:

```
if content_length == 0 or not self._raw:
```

Perhaps this would be better:

```
if content_length == 0 or not self._raw.strip():
```

tobixen:
Nevermind, I think I misunderstood it.

aaujon:
Thanks for the quick merge :)

PR #301, opened 2023-03-08
Fix #300 : enable huge_tree in XMLParser to handle very big events

### LINKED ISSUE #300: Add support for very big event
Hi,

The current code doesn't work with events with very big content, exception I get is : 
`Traceback (most recent call last):
  File "/home/arnaud/.local/lib/python3.10/site-packages/caldav/davclient.py", line 72, in __init__
    self.tree = etree.XML(
  File "src/lxml/etree.pyx", line 3231, in lxml.etree.XML
  File "src/lxml/parser.pxi", line 1915, in lxml.etree._parseMemoryDocument
  File "src/lxml/parser.pxi", line 1803, in lxml.etree._parseDoc
  File "src/lxml/parser.pxi", line 1144, in lxml.etree._BaseParser._parseDoc
  File "src/lxml/parser.pxi", line 618, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 728, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 657, in lxml.etree._raiseParseError
  File "<string>", line 202389
lxml.etree.XMLSyntaxError: xmlSAX2Characters: huge text node, line 202389, column 65
`
This is produced when handling an event with a very big description, I manage to fix this by activating 'huge_tree' in XMLParser in davclient.py. This option is not activated by default du to a security issue : https://lxml.de/api/lxml.etree.XMLParser-class.html
I'm not sure exactly what it is about but we may want to enable it by default for caldav or at least add an option for it in DAVClient ?

aaujon:
A proposition to fix https://github.com/python-caldav/caldav/issues/300
Activating huge_tree in XMLParser allow to handle very big caldav events.


tests/test_caldav_unit.py
```diff
@@ -156,7 +156,7 @@ def MockedDAVResponse(text):
     resp.reason = "multistatus"
     resp.headers = {}
     resp.content = text
-    return DAVResponse(resp)
+    return DAVResponse(resp, False)
```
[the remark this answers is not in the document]
aaujon: you're right, I will fix that.

caldav/davclient.py
```diff
@@ -36,13 +36,15 @@ class DAVResponse:
     tree = None
     headers = {}
     status = 0
+    huge_tree = False
 
-    def __init__(self, response):
+    def __init__(self, response, huge_tree):
```
tobixen, line 41: huge_tree should be defaulted to False.  Also, ref my comment on #300, maybe a more "future proof" way of doing this is to pass the DAVClient object rather than huge_tree.

aaujon: Yes I noticed that from the last failed CI build.
I will rewrite that part and pass the DAVClient object, it will be easier to maintain and add new future options.

tobixen: Any progress?  Do you want me to take over?

aaujon: What do you think about this version ? Pro is it doesn't break anything if someone uses DAVRequest directly. If you want to tweak some changes, you can go ahead, I will do it otherwise.

tobixen: Looks good.  (I dislike a bit that the huge_tree-attribute is stored doubly up in the DAVResponse class, but ... yes, backward compatibility is important and this seems to be the simplest solution to that).

Unit tests or functional test would be good.

aaujon: I added a test, I don't like to add a 10MB file for that, maybe we can autogenerate a random content..
A bout the double huge_tree attribute, I didn't want to duplicate some logic code into XMLParser callbut I can add a function has_huge_tree in DAVResponse, so we dont have to store it twice, do you want me to do that ?

tobixen: About autogeneration ... it doesn't need to be so very random, some silly code like this would probably do:

```
event.icalendar_component['DESCRIPTION'] = "\n".join(["very long description - this is line %i" % x for x in range(0,1234567)
```

... or even ...

```
event.icalendar_component['DESCRIPTION'] = "This is a very long description\n" * 1234567
```

as for `self.has_huge_tree()` vs `self.huge_tree` ... do whatever seems easiest/best for you.

aaujon:
Tests are failing because rootdir is not the one I expected, fixed in new commit

tobixen:
"tox -e style", and then it's ready?

tobixen:
(the 80 character line length restriction in PEP-8 is a terrible constraint sometimes)

aaujon:
All right, shoud be good now.

aaujon:
Thanks for the review :)

tobixen:
I can't promise a quick release of caldav v1.3.0 this time, but hopefully in a week or two.

PR #306, opened 2023-05-22
add unit test and documentation for custom HTTP headers 

JasonSanDiego:
unit test and documentation related to https://github.com/python-caldav/caldav/issues/285


tobixen:
I think the test breakage is unrelated to this pull request, I'll look more into it.

JasonSanDiego:
This was my first time using tox. I spent a lot of time trying to get the docs test to work properly both before and after making the actual fix I was working on. Unfortunately, I could not make it work.

Issue #310, opened 2023-05-29
Error with calendar.date_search() and Google Calendar 


flozz:
I am using this library to fetch events of a meeting room calendar from Google Calendar.

When I try to search events by date I have the following error:

```
ERROR:caldav:Deviation from expectations found.  Please raise an issue at https://github.com/python-caldav/caldav/issues or reach out to t-caldav@tobixen.no, include this error and the traceback and tell what server you are using
Traceback (most recent call last):
  File "[...]/__env__/lib/python3.11/site-packages/caldav/lib/error.py", line 28, in assert_
    assert condition
AssertionError
Traceback (most recent call last):
  File "[...]/./gcal.py", line 78, in <module>
    events = calendar.date_search(start=start_date, end=end_date)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/__env__/lib/python3.11/site-packages/caldav/objects.py", line 907, in date_search
    objects = self.search(
              ^^^^^^^^^^^^
  File "[...]/__env__/lib/python3.11/site-packages/caldav/objects.py", line 1069, in search
    if any(key in component for key in recurrance_properties):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/__env__/lib/python3.11/site-packages/caldav/objects.py", line 1069, in <genexpr>
    if any(key in component for key in recurrance_properties):
           ^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

You will find a minimal example to reproduce this issue in the following article: https://blog.lasall.dev/post/tell-me-why-google-and-caldav/

tobixen:
Thanks for the report. 

I don't know when I will have time to look properly into it, but if you could dig a bit and eventually write a pull request, it would be much appreciated.  Here is the code section that fails (assuming you're on v.1.2.1):

```python
            for o in objects:
                ## This should not be needed
                o.load(only_if_unloaded=True)
                component = o.icalendar_component
                recurrance_properties = ["exdate", "exrule", "rdate", "rrule"]
                if any(key in component for key in recurrance_properties):
                    o.expand_rrule(start, end)
```

Apparently it breaks because `component` is `None`, while it's expected to be a icalendar object from the icalendar library.  The trivial work-around is to throw in something like this:

```
                if component is None:
                    continue
```

[...]

PR #312, opened 2023-06-08
Avoid crashing when object's component is missing

flozz:
Related issue: #310 

This PR avoid a crash when the VCALENDAR does not contains a VEVENT component.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -1070,6 +1070,8 @@ class Calendar(DAVObject):
                 ## This should not be needed
                 o.load(only_if_unloaded=True)
                 component = o.icalendar_component
+                if component is None:
+                    continue
                 recurrance_properties = ["exdate", "exrule", "rdate", "rrule"]
                 if any(key in component for key in recurrance_properties):
                     o.expand_rrule(start, end)


tobixen:
Sorry for the late reply.  You stated earlier that this workaround caused problems further downstream.  I think my proposed (but totally untested) workaround at https://github.com/python-caldav/caldav/issues/310#issuecomment-1568622879 may be better, as it filters out those empty responses rather than just ignoring them.

Anyway, whatever floats the boat - if this workaround suffices for you, it is better than the current code anyway.

flozz:
> Sorry for the late reply.

No problem :)

> You stated earlier that this workaround caused problems further downstream.

I only had some event with missing `vevent` object so I had to filter the event list with 

```python
events = [e for e in events if hasattr(e.instance, "vevent")]
```

> I think my proposed (but totally untested) workaround at https://github.com/python-caldav/caldav/issues/310#issuecomment-1568622879 may be better, as it filters out those empty responses rather than just ignoring them.

It is probably better. I do not have enough time to test currently, but I will try to test it when I have some time. Anyway it is a step in the right direction and we can improve it later.

Thank you for the merge :)

Issue #313, opened 2023-06-13
Get all calendars returns objects without ids


shikasta-net:
I've just run into this while trying to migrate issues from another task manager to a caldav one.  
When I create a calendar on the target server with `dav_client.principal().make_calendar`, the returned object has an `id` (whether I specify one or not). But if I try to look up all calendars on the target server with `dav_client.principal().calendars()` the returned lists of objects all have `id=None`.  The issue is in [`CalendarSet::calendars`](https://github.com/python-caldav/caldav/blob/a08d8396cb1ef824bd6f552d781be900769530ef/caldav/objects.py#L385)
where the data from the server doesn't seem to provide an id.  This seems weird.  I can to try to fix it, I'm just not sure if there's some design reason that it won't be possible.

tobixen:
> I've just run into this while trying to migrate issues from another task manager to a caldav one.

That's interesting, since I'm trying to do task management through my `plann` project.  IMHO, the icalendar standard is not really optimized for task management, I'm doing various workarounds in `plann`, it would be nice to have a look into other ical-based task managers to see how they get around the limitations.

> where the data from the server doesn't seem to provide an id.

The short answer is that the URL itself is the ID - so from that, I don't see any need of changing the code - though if you have any suggestions, I'd like to hear.

There is no well-defined concept of an ID here, and I'm sorry that the caldav library is referring to ID every here and there.   Ideally all references should be for "UID" and "URL" and nothing more than that - but I consider backward compatibility to be very important.

There are two standards here, it's CalDAV which is the communication part between the client and the server, and it's the icalendar, the actual calendar data.  In CalDAV, every resource is uniquely idenfitifed by an URL.  In the  python library, sometimes "ID" is used as the last part of this URL.

In the icalendar library, every event, journal entry or task has an UID field - so this is the unique identifier in the icalendar context.  The python caldav library attempts to map the UID of new events/tasks/journal entries into the URL.

shikasta-net:
Using the URL to populate the field was my workaround:
```
  for c in cal_dav_client.calendars():
    c.id = c.canonical_url.split('/')[-2]
    projects[c.id] = c
```
And what I'm thinking is that could be done inside `cal_dav_client.calendars()`.  This is maybe only an issue when migrating from one URL base to another; I'm trying to keep things simple by copying the already unique IDs for everything but the names had to change, so it then got weird to look up if a Calendar had already been created.  I'll see if I can add a minimal change that would populate the ID field from the URL.

[...]

PR #314, opened 2023-06-19
Unpack id when populating all calendar list

shikasta-net:
To address #313: lack of ID field populating in "all calendar" request.  I included fall-back in case the URL isn't as I expect.


tobixen:
Sorry, I forgot about this one

PR #315, opened 2023-07-16
add props parameter to search function

ge-lem:
Add the possibility to have extra props for search function.

My needs is to add `props=[dav.GetEtag()]` in the search.


tobixen:
Can you run `tox -e style`?

(yes, I can do it myself, but better you do it for a cleaner git history)

PR #323, opened 2023-07-20
Don't forget to set `.davclient` in `DAVResponse`

sobolevn:
`DAVResponse` already has class-level `davclient = None` set, looks like it is not set on instance-level by accident.

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -45,6 +45,7 @@ class DAVResponse:
         log.debug("response status: " + str(self.status))
 
         self._raw = response.content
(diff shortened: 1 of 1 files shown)


tobixen:
This classifies as a bugfix, though not a critical one since `DAVResponse.davclient` is not used anywhere.  It's related to pull request #300.  I think I will refactor a bit and replace `DAVResponse.huge_tree` with `DAVResponse.davclient.huge_tree`.

tobixen:
(hence I will not release any 1.3.7 today.  1.3.6 is still considered good)

PR #327, opened 2023-07-31
Fix OverflowError in caldav.elements.cdav._to_utc_date_string

tamarinvs19:
We are developing an open-source automatic test generation [tool](utbot.org) for Python. We have run the tool on a number of open source projects including your. We have generated a test that leads to an `OverflowError` in function ` caldav.elements.cdav._to_utc_date_string`.

This PR fixes this problem.

The first test is error suite before fix, the second - regression suite after fix.


tobixen:
The `caldav.elements.cdav._to_utc_date_string` has a `return ts.strftime("%Y%m%dT%H%M%SZ")`, so I find it strange that it returns `10101T000000Z` and not `00010101T000000Z`

tamarinvs19:
> Hm, test breaks now - the `caldav.elements.cdav._to_utc_date_string` has a `return ts.strftime("%Y%m%dT%H%M%SZ")`, so I find it strange that it returns `10101T000000Z` and not `00010101T000000Z`

Oh, it is a feature of linux CPython implementation. I rewrote regression test with `.strftime`.

tobixen:
> Oh, it is a feature of linux CPython implementation.

Or perhaps it's a bug? :-)

It's bad if different python implementations returns different things, so it should probably be reported upstream (but I don't have time doing any research into that now).

-- 
Tobias Brox
Senior System Consultant
Redpill Linpro AS - Changing the game
Mobil: +47 917 000 50 
Kontor: +47 215 441 68

Issue #328, opened 2023-08-12
Typehints should be implemented


tobixen:
ref https://github.com/python/typeshed/pull/10485#issuecomment-1643745094

(Any contributions welcome, it's not needed to cover everything in one pull request)

tobixen:
Sure

PR #331, opened 2023-08-18
Typehints

mahnazfallahh:
i hope this message finds you well , i add typehints in base.py and cdav.py and I'm open to any feedback .

### DIFF
--- a/caldav/elements/base.py
+++ b/caldav/elements/base.py
@@ -3,6 +3,7 @@
 from caldav.lib.namespace import nsmap
 from caldav.lib.python_utilities import to_unicode
 from lxml import etree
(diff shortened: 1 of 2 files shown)


tobixen:
(This is a partial implementation of #328)

Apparently one should use square brackets rather than ordinary brackets, ref https://docs.python.org/3/library/typing.html#typing.Union

Be aware that  most of the work is already done in the [typeshed project](https://github.com/python/typeshed/tree/main/stubs/caldav/caldav), code can be merged from there - see also https://github.com/python/typeshed/pull/10485#issuecomment-1643745094

mahnazfallahh:
ok i'll fix them

PR #333, opened 2023-09-19
set_related: allow custom types with set_reversed=False

Zocker1999NET:
Method failed before when using a custom relation type. Specification explicitly allows custom X-Types for non-standardized use cases. This also makes the library future-proof if futher relation types are added.

Also extract RELTYPE_REVERSER constant from method for good practice.

Tested for my use case (using a custom type) locally.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -31,7 +31,7 @@ except ImportError:
     from urllib import unquote, quote
 
 try:
-    from typing import Union, Optional
+    from typing import ClassVar, Union, Optional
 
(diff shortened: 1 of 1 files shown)


tobixen:
Test breakages on 3.7 and 3.8 does not seem to be related to this patch

Zocker1999NET:
Thanks for the fast answer & merge

tobixen:
@Zocker1999NET, do you by any chance have the time to comment on my related work in #336 ?

Issue #332, opened 2023-09-06
Tests are broken - TypeError from dulwich


tobixen:
Waiting for feedback from upstream, ref https://github.com/jelmer/xandikos/issues/261

tobixen:
still a problem with github tests running python 3.8 and 3.7

tobixen:
(this only affects the tests - the caldav library should still work on python 3.7 and python 3.8.  We need to hand-pick a working version/commit of the xandikos library when doing tests for 3.7 and 3.8 I suppose.  I will look into it ... before the next release).

PR #335, opened 2023-09-19
work around icalendar ignoring additional parameters

### LINKED ISSUE #334: set_relation reltype can be ignored because of type(uid)==icalendar.vText
## Description

When using `set_relation` on a component (e.g. todo), the reltype might be ignored. This is because the library [icalendar](https://github.com/collective/icalendar/issues) ignores properties if the value of a property is already of an internal type (see [code in question](https://github.com/collective/icalendar/blob/df769f88694745d3babb87457bc3080965aa62c4/src/icalendar/cal.py#L139-L141)). This happens if the UID is directly taken from the icalendar object and is of type `vText`.

## Testing

```python3
cal: caldav.Calendar  # assume valid calendar
task = cal.add_todo(summary="First Task")
dependent = cal.add_todo(summary="Dependent")
dependent.set_relation(other=task, reltype="DEPENDS-ON", set_reverse=False)
print(dependent.icalendar_instance.to_ical())
```

Check output that reltype got lost.

## Possible Mitigations

Converting the uid to a string explicitly before calling `self.icalendar_component.add` in `set_relation` works for me. This may be a possible workaround until icalendar potentially fixes this unexpected behavior. I can a add a PR for that, if requested.

I also submitted this issue to icalendar (see collective/icalendar#557). However, I see that changing this behavior in icalendar might introduce issues for other dependent projects, so I do not expect them to fix this issue soon.

Zocker1999NET:
- workarounds https://github.com/collective/icalendar/issues/557
- fixes #334 
  - see issue description for reasoning

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -1840,6 +1840,12 @@ class CalendarObjectResource(DAVObject):
             if rel == uid:
                 return
 
+        # without str(…), icalendar ignores properties
+        #  because if type(uid) == vText
+        #  then Component._encode does miss adding properties
+        #  see https://github.com/collective/icalendar/issues/557
+        #  workaround should be safe to remove if issue gets fixed
+        uid = str(uid)
         self.icalendar_component.add(
             "related-to", uid, parameters={"RELTYPE": reltype}, encode=True
         )


tobixen:
Thanks, excellent.

The problems with broken tests are related to #332 which was supposedly fixed upstream.

PR #338, opened 2023-10-03
Remove unnecessary use of a regex

scop:


tobixen:
Thanks for this, too! :-)

PR #339, opened 2023-10-05
Fix missing icalendar_component

e-katkov:
We faced the problem that icalendar_component missing in some cases. Added checks if the event has the icalendar_component attribute.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -1470,9 +1470,10 @@ class Calendar(DAVObject):
             ## but at one point it broke due to an extra CR in the data.
             ## Usage of the icalendar library increases readability and
             ## reliability
-            item_uid = item.icalendar_component.get("UID", None)
-            if item_uid and item_uid == uid:
-                items_found2.append(item)
+            if item.icalendar_component:
+                item_uid = item.icalendar_component.get("UID", None)
+                if item_uid and item_uid == uid:
+                    items_found2.append(item)
         if not items_found2:
             raise error.NotFoundError("%s not found on server" % uid)
         error.assert_(len(items_found2) == 1)


tobixen:
That means the server hasn't returned valid data.  Does it work as intended for you with this patch?  Ideally I'd like to have a test case and some credentials to the server for testing, if possible.

In any case, my intention is for the caldav library to be pragmatic and work over as many caldav servers as possible, even when they are breaking the standards, so I will eventually merge this request.

tobixen:
(test failures for 3.7 and 3.8 are due to problems with the testing framework itself, nothing to bother with - I haven't gotten around to look into it yet).

e-katkov:
> That means the server hasn't returned valid data.

We have had "empty" events received in some cases. Such events had VERSION, PRODID, TZID, DTSTART, TZNAME and similar fields. In addition, they did not contain the UID, SUMMARY, DESCRIPTION, ORGANIZER, ATTENDEE, LAST-MODIFIED, RRULE and similar fields.

> Does it work as intended for you with this patch?

Yeah. That's what made it work.

> Ideally I'd like to have a test case and some credentials to the server for testing, if possible.

Unfortunately, that's not possible. We have only gotten this for one user so far. It is not clear how "empty" events are created.

tobixen:
That saves me from spending time doing research on this at least :-)  (may I ask what calendar server it is?)

e-katkov:
Google calendar)

e-katkov:
May I ask when you plan to update version that will have these changes?

tobixen:
sorry that the "worst case" has slipped a bit, hope to get it out soon.

PR #347, opened 2023-11-26
Initialize stamped and ended, implementing the small optimization todo

yuwash:
The lack of initialization of the `stamped` variable was causing a failure from an unexpected place when running it with a faulty ical that lacked a `BEGIN:V…` before having the first `DTSTAMP`. Probably the same with `ended`. Only affects bad strings, so it’s more of an error handling improvement…

### DIFF
--- a/caldav/lib/vcal.py
+++ b/caldav/lib/vcal.py
@@ -121,6 +121,11 @@ class LineFilterDiscardingDuplicates:
     This must be called line by line in order on the complete text, at
     least comprising the complete vobject.
     """
+
+    def __init__(self):
+        self.stamped = 0
+        self.ended = 0
+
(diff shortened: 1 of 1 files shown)


tobixen:
Thanks for the contribution.  Unfortunately I don't have time to do a code review right now, but possibly later today or tomorrow :-)

tobixen:
It would be nice with ...

1) test code proving that the current code is breaking

2) unit test proving that the class does what it's intended to do

tobixen:
And 3 also, updates to the `CHANGELOG.md`-file.  Anyway, if this actually solves a bug, I suppose it's best to just merge it now and indicate in a separate issue that we need tests.

PR #363, opened 2023-12-19
Remove dependency on pytz

### LINKED ISSUE #231: Remove dependency on pytz
The pytz and tzlocal libraries are deprecated, in modern python versions timezone handling is included in the embedded libraries - so we should remove all dependencies on pytz.  I think I sort of did this already, but then I got into problems.  The task is basically to look into the code and search for all instances of pytz and tzlocal and ensure that it's either removed or only used as a fallback should timezones not be available.  I vaguely remember that I already did this but had to comment out the work due to problems with the icalendar library.

Problems:

* The icalendar library had problems with the new timezone objects.  While waiting for https://github.com/collective/icalendar/issues/333 to be fixed and released it was needed to stick to pytz.
* We need the library to be backwards-compatible.  For python version 3.x this may be fixed by adding a dependency on backports.timezone - see https://pypi.org/project/backports.zoneinfo/.  With python2, I guess we still has to rely on pytz. Arguably, we should drop support for python2, but I have some comments on that at https://github.com/python-caldav/caldav/pull/228.

sim0nx:
- Removed dependency on pytz
- kept tzlocal only for testing
- added backports.zoneinfo in case of <py3.9 for tests

fixes #231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project should more or less adhere to [Semantic Versioning](https://semver.
 ## [x.x.x] - unreleased
 
 * Initial work at integrating typing information. Details in https://github.com/python-caldav/caldav/pull/358
+* Remove dependency on pytz. Details in https://github.com/python-caldav/caldav/issues/231
 
 ## [1.3.9] - 2023-12-12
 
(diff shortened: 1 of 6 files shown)


tobixen:
awesome

PR #364, opened 2023-12-19
Use setuptools-scm / pyproject.toml (modern packaging)

### LINKED ISSUE #357: Look into setuptools-scm
When doing releases, only files tracked in git and all the files tracked in git should be included in the release posted to pypi.

We should look into https://pypi.org/project/setuptools-scm/ - that plugin is supposed to help with this.

sim0nx:
- migrate setup.py to pyproject.toml
- use setuptools-scm

fixes #357

Please note that in order to build packages, you should after merging this, do the following:
```
pip install -U pip build
python -m build
```

The __version__ variable will be available just as before, but auto-generated.

Ref: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives


tobixen:
> Please note that in order to build packages, you should after merging this, do the following:
> 
> ```
> pip install -U pip build
> python -m build
> ```

This should probably be included in the documentation

sim0nx:
> > Please note that in order to build packages, you should after merging this, do the following:
> > ```
> > pip install -U pip build
> > python -m build
> > ```
> 
> This should probably be included in the documentation

updated release-howto.md

tobixen:
Another thing, the `RELEASE-HOWTO.md` probably needs some modifications (after several really silly embarrassing mistakes while releasing, I've made a rather complicated procedure to make sure I'm on the correct branch, that the version number is updated, that the tag is correct and that I'm doing the dist from a clean checkout - I suppose some of the steps may be sidestepped now).

tobixen:
Seems like my last comment was posted some seconds too late :-)

tobixen:
Or maybe not.  There are still some steps in the RELEASE-HOWTO that doesn't make sense now I think (like grepping for the version number from `__init__.py`)

sim0nx:
I removed those obsolete parts. Should hopefully be fine now :-)

tobixen:
```python
>>> import caldav
>>> caldav.__version__
'1.3.10.dev4+g70e7856.d20231219'
```

lovely :-)

PR #366, opened 2023-12-26
fix: google may return empty ical

zhwei:
something like:

```
BEGIN:VCALENDAR
PRODID:-//Google Inc//Google Calendar 70.9054//EN
VERSION:2.0
CALSCALE:GREGORIAN
X-WR-CALNAME:TEST
X-WR-TIMEZONE:Asia/Shanghai
BEGIN:VTIMEZONE
TZID:Asia/Shanghai
X-LIC-LOCATION:Asia/Shanghai
BEGIN:STANDARD
TZOFFSETFROM:+0800
TZOFFSETTO:+0800
TZNAME:CST
DTSTART:19700101T000000
END:STANDARD
END:VTIMEZONE
END:VCALENDAR
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ This project should more or less adhere to [Semantic Versioning](https://semver.
 * Georges Toth (@sim0nx) did a lot of efforts lifting up the project to more modern standards.
 * A hook for collecting debug information has been in the pull request for ages.  I've decided to include it in 1.4.0.
 * Code formatting / style fixes.
+* Search method did some logic handling non-conformant servers (loading data from the server if the search response didn't include the icalendar data, ignoring trash from the Google server when it returns data without a VTODO/VEVENT/VJOURNAL component.
+
+### Fixed
+
+* Revisited a problem that Google sometimes delivers junk when doing searches - credits to github user @zhwei in https://github.com/python-caldav/caldav/pull/366
+* There were some compatibility-logic loading objects if the server does not deliver icalendar data (as it's suppsoed to do according to the RFC), but only if passing the `expand`-flag to the `search`-method.  Fixed that it loads regardless of weather `expand` is set or not.  Also in https://github.com/python-caldav/caldav/pull/366
+
+### Changed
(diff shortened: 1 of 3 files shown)


tobixen:
Please use `o.is_loaded()` to check if the object has a payload or not, it's more efficient.

I remember that this issue (google returning empty objects) have been looked into a couple of times, but I don't remember the conclusion.  Will `o.load()` help?

tobixen:
I'm a bit baffled why the tests break.  I'm looking into it.

tobixen:
I do not like the usage of `o.icalendar_component` here, becuase it may involve converting the object into an object from the icalendar library.  Anyway, I can see that it's already used in the same function, after `if kwargs.get("expand", False):`, so then my objection is moot.  :-)  I should probably go over and refactor at a later stage.  It's still needed to figure out why the test breaks though.

zhwei:
@tobixen  Thanks for your review.

> Will o.load() help?

It has no effect, the original data have been loaded, but have no event/todo/etc.

FYI:
this event looks like a event's recurrence item, id format is `idexample001_R20231222@google.com`, i cannot reproduce this event, but my user have 😞 . 
The weird thing is i call `search` with expand=False, this recurrence item should not return standalone. If do not pass sort_keys parameter, the return result is like

```
Event(https://...../idexample001@google.com.ics)
Event(https://...../idexample001_R20231222@google.com.ics)
```

tobixen:
I promised to look into why the test breaks - but then I got interrupted and busy.  I will try and have a look now.

tobixen:
Sorry for overwriting your changes.  I think it will be better this way.  I think the special logic for dealing with objects that aren't loaded or doesn't contain components is not specific to when `expand=True`, and hence it should be outside the if.

I consider this to be a bugfix, as it will cause consistent and expected behaviour even when the caldav server doesn't conform to the RFCs.

tobixen:
Could you confirm that this solves your problems?

zhwei:
@tobixen Yes, it's works for me.

Issue #391, opened 2024-04-13
Change default User-Agent?


tobixen:
Also ref https://github.com/home-assistant/core/issues/105292

The default User-Agent is set to `Mozilla`, I'm considering to change it to `python-caldav` or something like that. 

The Mozilla default was set by @cyrilrbt  in 2010 or earlier.  Was there any reason for choosing Mozilla rather than something else?  Also, could there be caldav servers out there that behaves differently dependent on the User-Agent header? Even the most innocent changes in this code has caused breakages on weird servers and inbound bug reports - and I've seen systems blocking traffic based on the User-Agent.  Perhaps it's best not to jinx the default setting?

cyrilrbt:
IIRC I set it to mozilla because some service was blocking me without a user agent, so this might be a breaking change.

PR #392, opened 2024-04-13
CHange default User-Agent from Mozilla/5 to python-caldav/$version

tobixen:
Resolves https://github.com/python-caldav/caldav/issues/391


tobixen:
I will leave this here for a while, in case someone wants to comment #391

Issue #394, opened 2024-04-29
Recurrent event with an exception instance leads to multiple `RECURRENCE-ID` values in the exception instance


dozed:
I created an iCalendar object as follow:
- Create bi-weekly event
- Edit the title of the second instance at 2024-04-25 in Thunderbird

This leads to an iCalendar object with two events, one master and one exception instance:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Mozilla.org/NONSGML Mozilla Calendar V1.1//EN
BEGIN:VEVENT
UID:c26921f4-0653-11ef-b756-58ce2a14e2e5
DTSTART;VALUE=DATE:20240411
DTEND;VALUE=DATE:20240412
DTSTAMP:20240429T181103Z
LAST-MODIFIED:20240429T181103Z
RRULE:FREQ=WEEKLY;INTERVAL=2
SEQUENCE:1
SUMMARY:Test 1
X-MOZ-GENERATION:1
END:VEVENT
BEGIN:VEVENT
UID:c26921f4-0653-11ef-b756-58ce2a14e2e5
RECURRENCE-ID;VALUE=DATE:20240425
DTSTART;VALUE=DATE:20240425
DTEND;VALUE=DATE:20240426
CREATED:20240429T181031Z
DTSTAMP:20240429T181103Z
LAST-MODIFIED:20240429T181103Z
SEQUENCE:1
SUMMARY:Test 1 (edited)
X-MOZ-GENERATION:1
END:VEVENT
END:VCALENDAR
```

When searching for the event via `caldav`, two instances are returned. The second instance contains the `RECURRENCE-ID` field twice as shown in the following test case:

```python
def testRecurringDateWithExceptionSearch(self):
    self.skip_on_compatibility_flag("read_only")
    self.skip_on_compatibility_flag("no_recurring")
    c = self._fixCalendar()

    # evr2 is a bi-weekly event starting 2024-04-11
    e = c.save_event(evr2)

    r = c.search(
        start=datetime(2024, 3, 31, 0, 0),
        end=datetime(2024, 5, 4, 0, 0, 0),
        event=True,
        expand=True,
    )

    assert len(r) == 2

    assert 'RRULE' not in r[0].data
    assert 'RRULE' not in r[1].data

    assert isinstance(r[0].icalendar_component['RECURRENCE-ID'], icalendar.vDDDTypes)
    # OK:
[...]

PR #395, opened 2024-04-30
Fixed duplicate RECURRENCE-ID when expanding events

dozed:
Fixes https://github.com/python-caldav/caldav/issues/394

There is an issue with Xandikos: when searching for the events a `500 Internal Server Error` is returned. This might need some investigation.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -2003,7 +2003,8 @@ class CalendarObjectResource(DAVObject):
         calendar = self.icalendar_instance
         calendar.subcomponents = []
         for occurrence in recurrings:
-            occurrence.add("RECURRENCE-ID", occurrence.get("DTSTART"))
+            if "RECURRENCE-ID" not in occurrence:
+                occurrence.add("RECURRENCE-ID", occurrence.get("DTSTART"))
             calendar.add_component(occurrence)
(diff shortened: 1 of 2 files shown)


tobixen:
```
  File "/home/runner/work/caldav/caldav/.tox/py/lib/python3.10/site-packages/xandikos/icalendar.py", line 1039, in asutc
153
    return dt.astimezone(pytz.utc).replace(tzinfo=None)
154
AttributeError: 'datetime.date' object has no attribute 'astimezone'
155
```

This seems to be a problem with Xandikos.

tobixen:
@jelmer - do you have time to look into the broken test here?

tobixen:
Sorry for forgetting about this one for so long.  Now I can't figure out how to rerun the tests.  Hm.

tobixen:
So the xandikos-problem is with your test code, it is using dates for DTSTART, while datetime probably would be more appropriate.  At the other hand, date is a valid type for DTSTART, so perhaps the problem is with xandikos.  Hm.

tobixen:
I have cherry-picked the commit and worked around the xandikos issues.  I will raise an issue on the xandikos issue tracker about the breakage.

PR #405, opened 2024-06-12
by default dict loops yield only key

ArtemIsmagilov:
https://realpython.com/iterate-through-dictionary-python/


tobixen:
thanks

PR #406, opened 2024-06-12
try use new versions github-actions and drop unsupported python3.7

ArtemIsmagilov:
Not sure if there is any real benefit from this update. Only if your tests are very long.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,16 +14,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python:
-          - '3.12'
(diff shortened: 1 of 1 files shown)


tobixen:
No idea about the github actions, probably smart to keep it on the latest versions ... but at the other hand, I have a policy of supporting EOL'ed python versions for a while.  Old python versions usually hang around in the wild for quite a while, and even though they may be "unsupported" by the Python Software Fundation, they may still be found in LTS-distros and supported by Canonical, Red Hat, etc.

ArtemIsmagilov:
the reason is great, let's leave py37 then

PR #408, opened 2024-06-12
change mut arg on None and added checker is None

ArtemIsmagilov:
https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil


ArtemIsmagilov:
Thank you)

PR #409, opened 2024-06-12
copy resolve mut arg in funcs, methods

ArtemIsmagilov:
this PR copy resolver problem #408

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -50,7 +50,7 @@ class DAVResponse:
     raw = ""
     reason: str = ""
     tree: Optional[_Element] = None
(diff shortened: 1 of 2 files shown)


ArtemIsmagilov:
You can, of course, strictly rewrite. At your discretion
```python
if arg is None:
    arg = {}
```

tobixen:
I think it's fine.

In most cases the dicts and lists passed as parameters are supposed to be read-only, then having `{}` or `[]` as default value isn't harmful.  At the other hand, there is no way to enforce or even indicate that objects passed as parameters shouldn't be modified by a method.

PR #411, opened 2024-06-12
remove unused line `---` in .pre-commit-config.yaml config file

ArtemIsmagilov:
removed unused line

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,3 @@
----
 repos:
   - repo: https://github.com/asottile/reorder_python_imports
     rev: v3.12.0


tobixen:
Some yaml linters demands that all yaml files starts with `----`

https://stackoverflow.com/questions/50788277/why-3-dashes-hyphen-in-yaml-file

So it should stay as it is, I think :-)

tobixen:
Maybe this can be removed?  Good to have a standard, but sometimes I like to annotate imports wih some `# comments`.   This automatic reordering of imports causes the comments to be moved away from the imports.

tobixen:
Aha ...

> The style chosen by reorder-python-imports has a single aim: reduce merge conflicts.

So we'll keep it like it is.

ArtemIsmagilov:
@tobixen I would certainly recommend isort to you. You can take a look https://pycqa.github.io/isort/

ArtemIsmagilov:
> Ага...
> 
> > Стиль, выбранный reorder-python-imports, преследует единственную цель: уменьшить конфликты слияния.
> 
> Поэтому мы оставим все как есть.

I think this is a small part of what isort does. However, like any tool, these are just the choice and preference of the developer. As a maintainer, it's up to you to decide.

tobixen:
I am relatively indifferent.  I don't know what's best from isort and the current sorting module.  I can see that the risk of merge conflicts when grouping the imports together on the same line exists, but it's not a big thing.  Another thing, when lots of things are to be imported, the 78 character limit gets in the way.

Perhaps I will change my mind at some point, but as for now I think we just keep it as it is.

ArtemIsmagilov:
I agree with you as long as you work as required. As we need something else, we will make new changes)

mbehrle:
> Another thing, when lots of things are to be imported, the 78 character limit gets in the way.

JFTR isort handles this.

PR #415, opened 2024-06-13
use isinstance with tuple classes

ArtemIsmagilov:
you can check some types with isinstance func


ArtemIsmagilov:
@tobixen, please, cancel action, this action is freeze

tobixen:
Probably a random fluke. All tests passed it says for me now at https://github.com/python-caldav/caldav/pull/415/checks

PR #416, opened 2024-06-13
Pep8 create func lambda vs def

ArtemIsmagilov:
https://peps.python.org/pep-0008/#programming-recommendations


ArtemIsmagilov:
autopep8 generate `def remove_bit(n): return n & (n - 1)`
in any case, it’s easier to correct the function if necessary)

ArtemIsmagilov:
you can cancel the PR. no difference. it just reads more clearly

tobixen:
Probably a random fluke.  All tests passed it says for me now at https://github.com/python-caldav/caldav/pull/415/checks

PR #417, opened 2024-06-13
refactoring maps code on set comprehension

ArtemIsmagilov:
1. bytecode faster
2. the code is easier
3. speed finding elements in set O (1) (hashable elemets)


tobixen:
Seems good to use set rather than list, but I will have to review this one a bit more carefully.

ArtemIsmagilov:
look example
```python
print(' sa asd asd\r\n \f \t dad as as , asd'.split())
['sa', 'asd', 'asd', 'dad', 'as', 'as', ',', 'asd']
```

ArtemIsmagilov:
This method is used only in 2 places to search for a string, set is suitable here. Map, filter and lambda are unnecessary here. For library users, nothing should change.

PR #422, opened 2024-06-13
try setting black on 120 length

ArtemIsmagilov:
the main thing here is setting up pyproject.toml. what do you think about the new string length?
is it comfortable for writing and reading code?
```text
...
# pyproject.toml
[tool.black]
line-length = 120

```

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -60,9 +60,7 @@ class DAVResponse:
     davclient = None
     huge_tree: bool = False
 
-    def __init__(
-        self, response: Response, davclient: Optional["DAVClient"] = None
-    ) -> None:
+    def __init__(self, response: Response, davclient: Optional["DAVClient"] = None) -> None:
         self.headers = response.headers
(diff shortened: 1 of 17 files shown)


tobixen:
https://stackoverflow.com/questions/88942/why-does-pep-8-specify-a-maximum-line-length-of-79-characters is an interesting read.

Personally I think the 79 character lmiit is stupid, for one thing, sometimes it makes it difficult to use tools like `grep` on the code, and it's quite silly when there is a long list in the code with one element pr line, and then the `black`-plugin goes in and convert one of the lines to a multiline.  Another argument is that the 79-character limit often makes me stop up in the middle of the coding just to spend efforts considering how to wrap the lines (at least that argument is moot since `black` does it for me now).  At he other hand, I see no reason why the `caldav`-project should deviate from the standard given in pep8.  I also think it's stupid to first format the code with a 79 character line limit, and then reformat it to 120 lines later.

ArtemIsmagilov:
Pep8 my favourite rules)

PR #427, opened 2024-07-11
feat: another example of using

Truenya:
just yet another example(found on stackoverflow)

### DIFF
--- a/examples/get_events_example.py
+++ b/examples/get_events_example.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+import json
+from os import environ as env
+
+import caldav
+
+username = env["CALDAV_USERNAME"]
+password = env["CALDAV_PASSWORD"]
+url = env["CALDAV_URL"]
+caldav_url = f"https://{url}/{username}/"
+headers = {}
+
+
+def fetch_and_print():
+    with caldav.DAVClient(
+        url=caldav_url,
+        username=username,
+        password=password,
+        # Optional parameter to set HTTP headers on each request if needed
+        headers=headers,
+    ) as client:
+        print_calendars_demo(client.principal().calendars())
+
+
+def print_calendars_demo(calendars):
+    if not calendars:
+        return
+    events = []
+    for calendar in calendars:
+        for event in calendar.events():
+            for component in event.icalendar_instance.walk():
+                if component.name != "VEVENT":
+                    continue
+                events.append(fill_event(component, calendar))
+    print(json.dumps(events, indent=2, ensure_ascii=False))
+
+
+def fill_event(component, calendar) -> dict[str, str]:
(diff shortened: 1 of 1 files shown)


examples/caldav-get-lib.py
```diff
@@ -0,0 +1,47 @@
...
+    ) as client:
+        print_calendars_demo(client.principal().calendars())
+
+
+def print_calendars_demo(calendars):
+    if not calendars:
+        return
+    events = []
+    for calendar in calendars:
+        for eventraw in calendar.events():
```
tobixen, line 29: `eventraw`  here is an object of class `Event`.  I wouldn't call it `raw`.

examples/caldav-get-lib.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+from icalendar import Calendar
```
tobixen, line 2: This import can go away (see comments further below) (and hence the "external library" is not so external anymore - the caldav library already has built-in support for icalendar objects!)

examples/caldav-get-lib.py
```diff
@@ -0,0 +1,47 @@
...
+        print_calendars_demo(client.principal().calendars())
+
+
+def print_calendars_demo(calendars):
+    if not calendars:
+        return
+    events = []
+    for calendar in calendars:
+        for eventraw in calendar.events():
+            for component in Calendar.from_ical(eventraw._data).walk():
```
tobixen, line 30: Use rather `eventraw.icalendar_instance.walk()`.

Also, `eventraw._data` is internal and not supposed to be used directly.  `eventraw.data` is more appropriate.

Tidbit: the icalendar instance created through `eventraw.icalendar_instance` is still linked with the `eventraw` object - so if you edit the icalendar instance and then later call `eventraw.data`, then it will include the changes done on the icalendar object.

tobixen (changes requested):
The icalendar library is not so external, the caldav library already has support for it.  Anyway, fix those issues and I'll include this example.

tobixen:
Also, style test breaks due to the pesky requirements that lines should not exceed 78 characters.  This can be fixed easily by running `tox -e style` and commit.

examples/caldav-get-lib.py
Truenya: fixed

examples/caldav-get-lib.py
Truenya: done

examples/caldav-get-lib.py
Truenya: done

tobixen:
Style check is still failing

Truenya:
@tobixen fixed

tobixen (approved):

PR #431, opened 2024-08-03
Clarify python 3.7-3.12

ArtemIsmagilov:
Extend classifiers , information python available versions


tobixen:
Oh ... we've dropped support for Python 3.6?  Well, whatever.

PR #433, opened 2024-08-13
Allow to reverse the sorting order on search function 

### LINKED ISSUE #432: Allow to reverse the sorting order on search function
For my particular use case, I need to sort events from newer to older. That is currently impossible as using `dtstart` sorts from older to new.

Luckily, adding an argument to the `search` function to reverse the order was quite easy.

I also added changes to the test suite. I couldn't run it as the test was skipped, but it should work. I tested the change within my project and worked just fine. It's not a big change.

twissell-:
Closes #432

There are some import order changes, I'm assuming that was tox as it has nothing to do with what I've done.


tobixen:
thanks

Issue #426, opened 2024-06-26
Make logging configurable through environment variables


tobixen:
The caldav library does produce log noise when used towards popular but non-conforming calendar servers.  There is a wish to mute such logging i.e. by setting up environmental variables for it.  See also https://github.com/home-assistant/core/issues/111633

MrEbbinghaus:
(This also comes from Home Assistant as well.)

Logging the messages as 'Error' seems inappropriate, with the message _This is probably harmless, ..._ being logged after the lib has detected and fixed a problem. 

Also, to avoid confusing log entries without a message, the diff log should either always be at debug level, or even better, included in the previous log. 
https://github.com/python-caldav/caldav/blob/4c381834953b0d64a2afbc0ad0b6268670e56883/caldav/lib/vcal.py#L108-L114

tobixen:
Great.  Would you mind creating a pull-request?

Gloomyeye:
Great to here that the logging will be better. It would be nice to change the logging object in the HA log entry from root to a caldav specific logging object of possible.

PR #438, opened 2024-10-06
Change fixup log lvl to warning and merge diff log messages into related parent log

MrEbbinghaus:
This PR merges the two fix up log messages into one. The diff log can be confusing when the log is viewed outside the console. (see image from Home Assistants log view)

image

~~I haven't touched the log level yet.~~

See: https://github.com/python-caldav/caldav/issues/426#issuecomment-2395409450


jelmer:
Those issues are fixed in xandikos master; I'll see if I can do a release today.

tobixen:
thanks

jelmer:
Released now

Issue #439, opened 2024-10-10
Test fails against radicale


smurfix:
```
$ pytest -sx tests/
tests/test_caldav.py
F

========================================== FAILURES ===========================================
_____________________________ TestScheduling.testInviteAndRespond _____________________________

self = ScheduleInbox(None), client = None
principal = Principal(http://dav.local/tester1/), url = None

    def __init__(
        self,
        client: Optional["DAVClient"] = None,
        principal: Optional[Principal] = None,
        url: Union[str, ParseResult, SplitResult, URL, None] = None,
    ) -> None:
        """
        Will locate the mbox if no url is given
        """
        super(ScheduleMailbox, self).__init__(client=client, url=url)
        self._items = None
        if not client and principal:
            self.client = principal.client
        if not principal and client:
            if self.client is None:
                raise ValueError("Unexpected value None for self.client")

            principal = self.client.principal
        if url is not None:
            if client is None:
                raise ValueError("Unexpected value None for client")

            self.url = client.url.join(URL.objectify(url))
        else:
            if principal is None:
                raise ValueError("Unexpected value None for principal")

            if self.client is None:
                raise ValueError("Unexpected value None for self.client")

            self.url = principal.url
            try:
                # we ignore the type here as this is defined in sub-classes only; require morechanges to
                # properly fix in a future revision
>               self.url = self.client.url.join(URL(self.get_property(self.findprop())))  # type: ignore

caldav/objects.py:1769:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
caldav/lib/url.py:181: in join
    pathAsString = str(path)
caldav/lib/url.py:106: in __str__
    return to_normal_str(self.__unicode__())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[ValueError('Unexpected value None for self.url_parsed') raised in repr()] URL object at 0x7f5d8919b170>

    def __unicode__(self) -> str:
        if self.url_raw is None:
            if self.url_parsed is None:
>               raise ValueError("Unexpected value None for self.url_parsed")
E               ValueError: Unexpected value None for self.url_parsed

[...]

PR #441, opened 2024-10-19
Make tests ignore broken server-side expand for radicale

tobixen:
updates https://github.com/python-caldav/caldav/issues/439


tobixen:
@jelmer - there seems to still be some breakage from xandikos here for python 3.8

jelmer:
That makes sense, since xandikos is python 3.9 and up only nowadays.

Issue #402, opened 2024-06-08
Server compatibility hints


tobixen:
This is something I have been considering for a long time - I'm a bit surprised that I haven't created an issue on it yet (or perhaps I forgot about it and didn't search well enough).

* `DAVClient` constructor should have a `server_compatibility_hints`, probably a dict, probably more or less built over the definitions in place under  `tests.compabiliiy_issues`.
* In some cases, the library should work around known compatibility-problems, for instance filtering searches on the client side. 
* In other cases, it should raise an error (perhaps a new ServerCompatibilityError) if it's to be expected that the server won't manage to fulfill a request.
* The current `tests.compabiliiy_issues` should be retired, and tests should honor the `server_compatibility_hints` instead.

(See also #401 and others)

tobixen:
I will mark this as "resolved".

Currently we're running a hybrid scheme with some old-style "incompatibility flags" (only used by the test framework for skipping various tests, no automatic checking for those flags) and a new "features" system where the structure/names of the features are a bit better thought through and where it's possible to store arbitrary amounts of information for each "feature" rather than just a boolean value, and where all those features can be automatically checked using my caldav-server-tester tool (albeit, the work there is currently neither merged to master nor released).

I wanted to eliminate the "compatibility flags" completely - but since starting to work with this (half a year ago, almost!)  I've totally spent more than 200 hours on the caldav project, most of this time has been spent fighting with compatibility issues.  I simply don't have time to eliminate that one completely.  I will close this task and create a new one for "mopping up the legacy incompatibility_issues list".

PR #451, opened 2024-11-18
Server checker

tobixen:
Creating a server checker that can output a list of "server quirks".

Some changes have been done to the test framework and even some to the actual code.  Should fix the CHANGELOG properly before merging this into the project.

Eventually, the "quirk list" should be organized a bit better, some of the quirks should be renamed, and I was also intending to move the server checker to a separate project https://github.com/tobixen/caldav-server-tester - that would probably be the better idea.

When this is done, I may start working on #402.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,19 +10,17 @@ This project should adhere to [Semantic Versioning](https://semver.org/spec/v2.0
 
 In version 2.0, the requests library will be replaced with niquests or httpx.  See https://github.com/python-caldav/caldav/issues/457.  Master branch is currently running niquests.
 
-(Also, in version 2.0, support for python 3.7 will be dropped, possibly also python 3.8.  Master branch supports both as for now).
+In version 2.0, support for python 3.7 will be dropped, possibly also python 3.8.  Master branch supports both as for now.
 
 ## [1.5.0] - [Unreleased]
-
-### Fixed
-
-* Readthedocs integration has been repaired (https://github.com/python-caldav/caldav/pull/453 - but eventually the fix was introduced directly in the master branch)
 
 ### Deprecated
 
 Python 3.7 is no longer tested - but it should work.  Please file a bug report if it doesn't work.  (Note that the caldav library pulls in many dependencies, and not all of them supports dead snakes).
 
 ### Changed
+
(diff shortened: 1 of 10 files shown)


niccokunzmann:
@tobixen, that seems to be related a bit to #463

A separate project could be an idea for that. Especially if it has a different focus. I am not to sure about this because I am new to caldav.

Also: For the Open Web Calendar in https://github.com/niccokunzmann/open-web-calendar/pull/660

- I am adding CalDAV support with this lib.
- I used the `responses` library to record API interactions AND create test cases with them. For claiming compatibility with edge-cases and quirks something like this could be useful for this lib, too.

Proposed Interface:

```python
client = DAVClient(...)
client.record_api()
# normal interactions
client.save_recorded_api("filename.yml")
```

Use an API recording:

```python
client = DAVClient(...)
client.load_recorded_api("filename.yml")
# all interactions here after will be done against the recorded API
client.stop_recorded_api()
```

This is really nice for testing, too and I am using this also in python-wise-banking-api.

We can also open another issue if you like to discuss this or have a call.

tobixen:
> @tobixen, that seems to be related a bit to #463

Didn't think of it, but you may be right.

> A separate project could be an idea for that. Especially if it has a different focus. I am not to sure about this because I am new to caldav.

Indeed, I already have on my list to move the code over to https://github.com/tobixen/caldav-server-tester 

> * I used the `responses` library to record API interactions AND create test cases with them. For claiming compatibility with edge-cases and quirks something like this could be useful for this lib, too.

Probably.  The work I've done here sort of works, but my intention was to create nice and readable code, and I feel I've failed miserably on that point.

With some luck I will be able to allocate some time during the weekend or later today for looking into those things.

niccokunzmann:
@tobixen, if you do not mind, I would add this kind of testing to my funded proposal by NLNet

tobixen:
Not sure exactly what it involves, but I'm ok with it :-)

tobixen:
I'm moving the script to the empty repository now.

My long-term plan is to allow the caldav library to probe the server to check for known compatibility-problems.  To avoid circular dependencies, I think the best would be to have the actual code doing the probing in the caldav library itself, and just offer a thin CLI as a separate package.  So my long-term plan is to move logic from the script and back into the caldav library.

The short-term plan is to wrap up the current script and make sure it can be installed from pypi and AUR, and run towards any server.

PR #452, opened 2024-11-20
use niquests

ArtemIsmagilov:
I would like to integrate this library. ll the advantages can be read here https://pypi.org/project/niquests/

What is the benefit of this caldav project?
- multiplexing (increasing productivity)
- support for asynchrony (you can add asynchronous interfaces)
- full compatibility with `requests`
- lots of functionality

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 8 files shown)


tobixen:
Making the caldav library async would be good, but I will have to read up a bit on this.

tobixen:
The part with doing requests (and particularly authentication) is actually quite fragile - much more fragile as it ought to be - there are many fragile server setups out there, I've had people filing bug reports almost every time I've been doing changes on it,  After reading and sleeping on this, I've concluded to take niquest into the master branch, but exclude it from any 1.x-releases.  I will make a 1.x-branch and backport all commits there that are not related or dependent on niquests.

This means the python 3.13-support, which is unrelated to nirequest, has to go into a separate pull request.

tobixen:
There is also the security aspect of depending on a new library with one maintainer/developer - the library haven't had too many eyeballs so far, and we've had some nasty supply-side attacks in the open source community throughout the last few years.  However, if would be very sad if nobody picks up niquest because niquest has "too few eyeballs".

ArtemIsmagilov:
Hi, great! et's do it like that!

To be honest, for some reason you use authentication yourself. Fortunately, there are already different classes and even arguments for different types of authentication. you should try using authentication out of the box.

tobixen:
Could you move the python 3.13 support over to a separate pull request?

tobixen:
Maybe the simplest is to cancel this pull request and make a new one with only one commit in it

tobixen:
All the white space changes in the toml file makes this complicated :-)

Issue #459, opened 2024-12-06
Forbidden, when request event detail with GET


pessimo:
I got a caldav server, and connect to it via apple calendar, it works fine with basic username and password auth. 
when i try to connect it via python-caldav, I can go through the auth part and get all the calendars. 

url:  https://caldav.feishu.cn/
method:  PROPFIND
headers:  {'User-Agent': 'python-caldav/1.4.0', 'Content-Type': 'text/xml', 'Accept': 'text/xml, text/calendar', 'Depth': '0'}
body: 
```xml
<?xml version=\'1.0\' encoding=\'utf-8\'?>\n<D:propfind xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav"><D:prop><D:current-user-principal/></D:prop></D:propfind>'
response:  <?xml version="1.0" encoding="UTF-8"?><D:multistatus xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav" xmlns:CS="http://calendarserver.org/ns/" xmlns:ICAL="http://apple.com/ns/ical/" xmlns:ME="http://me.com/_namespace/"><D:response><D:href>/</D:href><D:propstat><D:prop><D:current-user-principal><D:href>/u_ortn0136/</D:href></D:current-user-principal></D:prop><D:status>HTTP/1.1 200 OK</D:status></D:propstat></D:response></D:multistatus>
```

I can also get possibly the event_id via REPORT.

url:  https://caldav.feishu.cn/u_ortn0136/66A3A496-9B15-8003-66A3-A4969B158003/
method:  REPORT
headers:  {'User-Agent': 'python-caldav/1.4.0', 'Content-Type': 'application/xml; charset="utf-8"', 'Accept': 'text/xml, text/calendar', 'Depth': '1'}
body:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:"
	xmlns:C="urn:ietf:params:xml:ns:caldav"
	xmlns:CS="http://calendarserver.org/ns/&#34; xmlns:ICAL="http://apple.com/ns/ical/&#34; xmlns:ME="<url id="" type="url" status="" title="" wc="">http://me.com/_namespace/"><D:response>
	<D:href>/u_ortn0136/66A3A496-9B15-8003-66A3-A4969B158003/d787dfe9-4a60-41af-a610-fae0886ba7c0.ics</D:href>
	<D:propstat>
		<D:prop>
[...]

PR #460, opened 2024-12-06
Load by using a multiget

tobixen:
This should be a workaround for getting calendar information out from the Lark/Feishu systems - see #459 

It's not optimized for speed, more work on this later when we get the server quirk hints in place (ref #402)


tobixen:
No unit tests - we'd have to mock up too much.

No functional tests so far as this is only executed with lark and zimbra - but I'll probably fix that later as part of #402

PR #467, opened 2025-03-17
Fixing _set_icalendar_component: language not defined

mauritium:
`CalendarObjectResource`'s `_set_icalendar_component`, if the target had more than one subcomponent, would always fail because the variable `language` was not defined. I reckon this was a copy & paste bug. Contributing a simple fix with unit test.

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -2187,8 +2187,8 @@ class CalendarObjectResource(DAVObject):
             self.icalendar_instance.subcomponents[i[0]] = value
         else:
             my_instance = icalendar.Calendar()
-            my_instance.add("prodid", "-//python-caldav//caldav//" + language)
-            my_instance.add("version", "2.0")
+            my_instance.add("prodid", self.icalendar_instance["prodid"])
+            my_instance.add("version", self.icalendar_instance["version"])
             my_instance.add_component(value)
             self.icalendar_instance = my_instance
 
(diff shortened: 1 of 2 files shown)


tobixen:
Good catch.  I'm scratching my head a bit on this one.  Is this code section useful?  Would it be better to delete it?  If you see any use for the code, then your fix is a good one.  To get the pesky style test pass, please run `tox -e style`, `git commit --amend -a` and `git push -f`, and I'll merge it :-)

As for the cause, I suppose either some code lines were copied or moved.  The same code lines exists in `lib/vcal.py`, `create_ical`, in the latter the `language`-variable is set as parameter default.

mauritium:
indeed, I believe the code came from where someone tried to make a Danish imprint on the code ;-)

Style has been fixed, thanks for the help with git commands - hope this works now.

The code in question failed in (somewhat arcane) production code dealing with recurrence event exceptions on my end, hence the fix - even though I agree it probably caters for a fringe use case.

tobixen:
Ah, I didn't catch the small thing here that the code only fails when a object has more subcomponents.  That should be allowed.  We should support recurrences in the best possible way, without breakages.

Thanks.

(I did do an effort some years ago to ensure near full code test coverage, but that's some years ago ...)

tobixen:
> where someone tried to make a Danish imprint on the code ;-)

It doesn't have much to do with [Danification](https://denmarkification.com/) ... it's more like, we'd like to use English as language, but without using imperial units or weird date-formats (`en_IE` may do the trick there), but also without having to select one particular English-speaking nationstate.  I believe `en_DK` started as a joke, but there isn't much else to choose from.

PR #469, opened 2025-03-19
Skip nonexisting calendars

### LINKED ISSUE #468: Empty calendar generates exceptions
As described [here](https://github.com/home-assistant/core/issues/140798) we connect to our wiki calendar via caldav.
All is working fine, until one iterates through the calendars.

Somehow our wiki returns an empty calendar, resulting in `cal_id = ''` and `c_name = None`. This subsequently triggers an exception in `get_supported_components`.
The current implementation adds `cal_id = None` calendars in case of an exception to the list and also also empty `cal_id`s.

A simple `if cal_id == '': continue` @ [CalendarSet.calendars()](https://github.com/python-caldav/caldav/blob/master/caldav/objects.py#L456) fixes the issue for us, but I'm not deep enough in this library or caldav in general, if this approach would break anything else.
If this is viable, I could open a PR (but most likely an edit for this twoliner would be quicker).

soundstorm:
Fixes #468

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -454,6 +454,8 @@ class CalendarSet(DAVObject):
         for c_url, c_type, c_name in data:
             try:
                 cal_id = c_url.split("/")[-2]
+                if not cal_id:
+                    continue
             except:
                 log.error(f"Calendar {c_name} has unexpected url {c_url}")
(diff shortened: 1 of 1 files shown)


tobixen:
thanks.

PR #472, opened 2025-04-16
Servers that return a quoted URL in their path will now be parsed correctly

### LINKED ISSUE #471: principal() prints errors (but doesn't impact functionality) on Confluence Team‑Calendars because path comparison ignores percent‑encoding
Calling client.principal() against an Atlassian Confluence Team Calendars CalDAV endpoint logs an error which spams the log output because the library compares an unquoted path to percent‑encoded keys returned by the server:

```

[ 352:objects.py - get_properties ] Possibly the server has a path handling problem, possibly the URL configured is wrong.
Path expected: /plugins/servlet/team-calendars/caldav/principals/users/example@domain.com/, path found: ['/plugins/servlet/team-calendars/caldav/principals/users/example%40domain.com/']
Please consider raising an issue at https://github.com/python-caldav/caldav/issues or reach out to t-caldav@tobixen.no, include this error and the traceback and tell what server you are using.
Continuing, probably everything will be fine


```
### Steps to reproduce:
```python
from caldav import DAVClient

client = DAVClient(
    url="https://confluence.example.com/plugins/servlet/team-calendars/caldav",
    username="example@domain.com",
    password="********",
)
principal = client.principal()   # triggers the log above
```

### Environment:
| Item | Version |
|--------|--------|
| Caldav| 1.4.0 |
| Python | 3.10.11 |
| Atlassian Confluence | 7.19.17 |

## Proposed fix:

Allow for properties keys to be unquoted as well. Here is a working modification
Old code on DAVObject.get_properties:
```python
[...]
    path = unquote(self.url.path)
    if path.endswith("/"):
        exchange_path = path[:-1]
    else:
        exchange_path = path + "/"
    properties = {unquote(k): v for k, v in properties.items()}
    if path in properties:
        rc = properties[path]
[...]
```

Working code:
```python
[...]
    path = unquote(self.url.path)
    if path.endswith("/"):
        exchange_path = path[:-1]
    else:
        exchange_path = path + "/"
    properties = {unquote(k): v for k, v in properties.items()}  #### Unquote the paths on the properties keys, too
    if path in properties:
        rc = properties[path]
[...]
```

Thank you

edel-macias-cubix:
Fixes #471 
`pytest` looks the same before and after modifications
`tox -e style` has been run: no changes made

### DIFF
--- a/caldav/objects.py
+++ b/caldav/objects.py
@@ -312,6 +312,7 @@ class DAVObject:
             exchange_path = path[:-1]
         else:
             exchange_path = path + "/"
+        properties = {unquote(k): v for k, v in properties.items()}
 
         if path in properties:
             rc = properties[path]


edel-macias-cubix:
I tried to run the github actions workflow on my repo, which triggered another one here.
Putting this note here so it doesn't get executed twice after approval

tobixen:
Thanks

Issue #491, opened 2025-05-17
Robur issue with testCreateOverwriteDeleteEvent


tobixen:
`caldav/calendarobjectresource.py:634: KeyError`

href is given as `'/calendars/pythoncaldav-test/20010712T182145Z-123401%40example.com.ics'`

in the dict there is a `'/calendars/pythoncaldav-test/20010712T182145Z-123401@example.com.ics'`

Difference in encoding, `'@'` vs `'%40'`.

tobixen:
Problem is with `Event.load_by_multiget`.  Shouldn't this be utilizing `Calendar.calendar_multiget` instead?

We get this data from the server:

```
<?xml version="1.0" encoding="utf-8" ?>\n<D:multistatus xmlns:D="DAV:"><D:response><D:href>/calendars/pythoncaldav-test/20010712T182145Z-123401%40example.com.ics</D:href><D:status>HTTP/1.1 404 Not Found</D:status></D:response></D:multistatus>'
```

... but the code does not pick up the 404 properly.

I don't have time researching on how to fix this at the moment.

tobixen:
Tests seems to work for Robur again, using the #492 code.  I will leave the pull request open for a while, piggy-backing in whatever other fixes is needed for getting tests passing under GMX and Posteo.

PR #492, opened 2025-05-18
Pre-release v1.5 fixes

tobixen:
Including some real regression bugfixes.

Refers to #491, but sneaking in some other changes to the test framework and compatibility matrix.


tobixen:
I should look through the CHANGELOG also before committing.  And we should yank out niquests, that's slated for v2.0 rather than v1.5

Issue #482, opened 2025-05-12
Refactoring - `get_duration`, `get_due`, `get_dtend` to be obsoleted


tobixen:
The duality between duration and dtend/due is IMO an annoying misfeature of the icalendar standard, so I made some helper methods for dealing with this - but it doesn't belong to the caldav library, it belongs to the icalendar library.   Now we have the properties `duration`, `dtend` and `dtstart` available in the icalendar class.  We should rewrite the current logic to wrap the icalendar properties and make some deprecation warnings.

tobixen:
Claude suggests this is a 3-4 hours job - but it involves quite some rewriting, and I'm not sure it can be automated.  I'd say 6 hours.

tobixen:
Ref comments in the pull request, this has to wait not only for icalendar 7 to be released, but also for the icalendar 7 to be supported throughout the ecosystem.  So it is a v4.0-task, not a v3.0-task.

Also, the current pull-request is deprecating and delegating the logic in the current methods, but that's just half the job - it's also needed to go through all internal calls and rewrite it.

PR #498, opened 2025-05-25
Move start/end/duration logic to the icalendar library

tobixen:
Work on issue #482 - remove logic on `get_due`, `get_duration` etc, it belongs to the icalendar library.  Also references https://github.com/collective/icalendar/issues/496


tobixen:
This will have to go into 2.0 as the most recent versions of icalendar does not support python 3.8.

tobixen:
Procrastinating this until icalendar 7 is out.  It may be that it's needed to do more refactoring to support icalendar7.

tobixen:
This has to wait not only for icalendar7 to be released, but also for the whole ecosystem to support icalendar7.  So v4.0 probably.

PR #503, opened 2025-06-01
Documentation project - tutorial with doctests

tobixen:
I'd like to move stuff from the examples section to the documentation and run tests on it.  Manuel was suggested, so I will look into that.  And I was recommended to read on [Διάταξης](https://diataxis.fr/), so I've decided to focus on a tutorial.


tobixen:
This is nowhere near complete, but (probably?) much better than the doc that is out there

Issue #505, opened 2025-06-10
A bug in loading davobject


zealseeker:
https://github.com/python-caldav/caldav/blob/a6c39da205b419ea441820fe70e11f2444629f2d/caldav/calendarobjectresource.py#L625-L633

When exception is catched during `r = self.client.request(str(self.url))`, r is not assigned and `self.data = vcal.fix(r.raw)` will raise an excpetion.

Maybe we should move `self.data = vcal.fix(r.raw)` to the try block or simply `return self.load_by_multiget()` as originally in #475

tobixen:
I believe `return self.load_by_multiget()` is the correct thing here.  Let me check up a bit.

tobixen:
I've investigated and found:

* the `vcal.fix` is redundant, it's already done in the `data`-setter, so this logic should go away.
* It's not meant that the process should continue after the load_by_multiget - so it should be a return here.

Do you want to do a pull request, or should I just fix it?

zealseeker:
OK. I'll create a pull request.

PR #506, opened 2025-06-11
Fix a bug in loading davobject

zealseeker:
To solve the issue #505 

1. Removed `vcal.fix` as it has been implemented in _set_data
2. Return `load_by_multiget()` when GET fails.

### DIFF
--- a/caldav/calendarobjectresource.py
+++ b/caldav/calendarobjectresource.py
@@ -634,8 +634,7 @@ class CalendarObjectResource(DAVObject):
         except error.NotFoundError:
             raise
         except:
-            self.load_by_multiget()
-        self.data = vcal.fix(r.raw)
+            return self.load_by_multiget()
         if "Etag" in r.headers:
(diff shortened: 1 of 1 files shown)


tobixen:
You stil need to set `self.data=r.raw`.  

And also, does this work for you?  I ran the test code towards all the servers I had access to without hitting this problem, so I can't test this myself.  (What calendar server do you use?)

zealseeker:
@tobixen OK. I've added it. Since 'r' may not be assigned when an excpetion is catched, I add it in the try block. Please check whether it is fine.

I am using Lark (named Feishu in China) and have the same problem as #459

tobixen:
Right.  I was testing towards Lark at some point - but that was after my last change.

I think this looks good.

tobixen:
Some comments:

* With the `return`-statement now, code after the try-except-block should only run if there are no exceptions, so `r` will always be set.
* The next lines also depends on `r`

Hence it doesn't really matter if `self.data = r.raw` is inside or outside the try-block.

Issue #397, opened 2024-05-23
Recurrence instances gets stripped when doing `cal.object_by_uid(uid,comp_class=caldav.Event)`


tobixen:
Details in https://github.com/python-caldav/caldav/issues/35#issuecomment-2126481797

tobixen:
I believe this is some obscure server-side bug.  At least I was not able to reproduce towards Radicale nor Xandikos.  I've included the test code from @ptrba in the tests, I will try and run it towards the other servers I have access to.

tobixen:
I've been running the test towards all servers I had available without being able to reproduce the issue.

tobixen:
Closing as "completed" as I've added tests - but issue could not be reproduced.

PR #521, opened 2025-06-18
Test code for https://github.com/python-caldav/caldav/issues/397

tobixen:


tests/test_caldav.py
```diff
@@ -844,6 +844,51 @@ def testSchedulingMailboxes(self):
         inbox = self.principal.schedule_inbox()
         outbox = self.principal.schedule_outbox()
 
+    def testIssue397(self):
+        cal = self._fixCalendar()
+        cal.save_event(
+            """
+BEGIN:VCALENDAR
+VERSION:2.0
```
tobixen, line 852: Test fails with GMX because it's missing a PRODID.  DTSTAMP is also mandatory.

tobixen: I added the PRODI, but not the DTSTAMP

Issue #517, opened 2025-06-17
401 printed when getting a principal using username and password


edel-macias-cubix:
Hello!

When getting the principal, the first request made will always return error 401.
```python
with DAVClient(
        url=confluence_caldav_url, username="username", password="password"
) as client:
    my_principal = client.principal()
return my_principal
```

It seems that this behavior is caused by lines 477-486 in davclient.py.

The client is passed username and password, but if auth not provided/None, the first request will be fired without any authentication.
The caldav server returns a 401 error. In my case, the response is HTML, which then is printed to the console in line 141 of davclient.py

Functionality is not impacted but extra request are being sent to the caldav server and the logs get flooded.

If the fix is confirmed to be to assign auth a new HTTPBasicAuth object with the provided username/password, I can work on the fix and send a PR.
For now I just send the auth attribute instead of the username/password as a workaround.

Attached is the full output of the code above:
```
2025_06_17_15_02_14 WARNING  [  40:error.py             - weirdness                      ] Deviation from expectations found: Unexpected content type: text/html;charset=utf-8
2025_06_17_15_07_22 CRITICAL [ 141:davclient.py         - __init__                       ] Expected some valid XML from the server, but got this: 
b'<!doctype html><html lang="en"><head><title>HTTP Status 401 \xe2\x80\x93 Unauthorized</title><style type="text/css">body {font-family:Tahoma,Arial,sans-serif;} h1, h2, h3, b {color:white;background-color:#525D76;} h1 {font-size:22px;} h2 {font-size:16px;} h3 {font-size:14px;} p {font-size:12px;} a {color:black;} .line {height:1px;background-color:#525D76;border:none;}</style></head><body><h1>HTTP Status 401 \xe2\x80\x93 Unauthorized</h1><hr class="line" /><p><b>Type</b> Status Report</p><p><b>Description</b> The request has not been applied to the target resource because it lacks valid authentication credentials for that resource.</p><hr class="line" /><h3>Apache Tomcat/9.0.82</h3></body></html>'
Traceback (most recent call last):
  File "C:\Users\emacias\PycharmProjects\Halberto\.venv\Lib\site-packages\caldav\davclient.py", line 120, in __init__
    self.tree = etree.XML(
                ~~~~~~~~~^
        self._raw,
        ^^^^^^^^^^
    ...<2 lines>...
        ),
        ^^
    )
    ^
  File "src\\lxml\\etree.pyx", line 3286, in lxml.etree.XML
[...]

Issue #158, opened 2021-11-19
Unable to connect : 


Sigmun:
When I tried to connect to my nextcloud server on an OVH Web Host, I get the following error:

```
>>> client.principal()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in _update_chunk_length(self)
    696         try:
--> 697             self.chunk_left = int(line, 16)
    698         except ValueError:

ValueError: invalid literal for int() with base 16: b''

During handling of the above exception, another exception occurred:

InvalidChunkLength                        Traceback (most recent call last)
C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in _error_catcher(self)
    437             try:
--> 438                 yield
    439

C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in read_chunked(self, amt, decode_content)
    763             while True:
--> 764                 self._update_chunk_length()
    765                 if self.chunk_left == 0:

C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in _update_chunk_length(self)
    700             self.close()
--> 701             raise InvalidChunkLength(self, line)
    702

InvalidChunkLength: InvalidChunkLength(got length b'', 0 bytes read)

During handling of the above exception, another exception occurred:

ProtocolError                             Traceback (most recent call last)
C:\ProgramData\Anaconda3\lib\site-packages\requests\models.py in generate()
    757                 try:
--> 758                     for chunk in self.raw.stream(chunk_size, decode_content=True):
    759                         yield chunk

C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in stream(self, amt, decode_content)
    571         if self.chunked and self.supports_chunked_reads():
--> 572             for line in self.read_chunked(amt, decode_content=decode_content):
    573                 yield line

C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in read_chunked(self, amt, decode_content)
    792             if self._original_response:
--> 793                 self._original_response.close()
    794

C:\ProgramData\Anaconda3\lib\contextlib.py in __exit__(self, typ, value, traceback)
    136             try:
--> 137                 self.gen.throw(typ, value, traceback)
    138             except StopIteration as exc:

C:\ProgramData\Anaconda3\lib\site-packages\urllib3\response.py in _error_catcher(self)
    454                 # This includes IncompleteRead.
[...]

PR #522, opened 2025-06-19
Log authentication warnings for servers returning HTML on 401.

edel-macias-cubix:
Log authentication warnings for servers returning HTML on 401 (behind proxies?), prompting users to use auth instead of username/password if the authentication method is known.
The warnings also print the servers available and compatible auth methods so the user can take action if they want to.

Improved DAVClient docstring to encourage auth parameter usage when no server auth probing is wanted/needed.

I decided upon logging a warning on why the logs are being printed and how to use the existing workarounds for this, because changing the current behavior could break something else.
Ref: https://github.com/python-caldav/caldav/issues/158#issuecomment-981891213

### DIFF
--- a/caldav/davclient.py
+++ b/caldav/davclient.py
@@ -462,15 +462,22 @@ class DAVClient:
         Parameters:
          * url: A fully qualified url: `scheme://user:pass@hostname:port`
          * proxy: A string defining a proxy server: `hostname:port`
+         * auth: A niquests.auth.AuthBase or requests.auth.AuthBase object. 
+          Use when the server's auth scheme is known.
+          Otherwise, pass username and password instead.
          * username and password should be passed as arguments or in the URL
          * auth, timeout and ssl_verify_cert are passed to niquests.request.
          * ssl_verify_cert can be the path of a CA-bundle or False.
          * huge_tree: boolean, enable XMLParser huge_tree to handle big events, beware
            of security issues, see : https://lxml.de/api/lxml.etree.XMLParser-class.html
 
-        The niquests library will honor a .netrc-file, if such a file exists
-        username and password may be omitted.  Known bug: .netrc is honored
-        even if a username/password is given, ref https://github.com/python-caldav/caldav/issues/206
+        The requests library will honor a .netrc-file, if such a file exists
+        username and password may be omitted.  Known bugs:
+         - .netrc is honored even if a username/password is given,
+           ref https://github.com/python-caldav/caldav/issues/206
+         - If the caldav server is behind a proxy or replies with html instead of xml 
+           when returning 401, warnings will be printed which might be unwanted.
+           Check auth parameter for details. 
         """
         headers = headers or {}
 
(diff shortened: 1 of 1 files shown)


edel-macias-cubix:
This hopefully resolves #517

caldav/davclient.py
```diff
@@ -462,15 +462,22 @@ def __init__(
...
          * auth, timeout and ssl_verify_cert are passed to niquests.request.
          * ssl_verify_cert can be the path of a CA-bundle or False.
          * huge_tree: boolean, enable XMLParser huge_tree to handle big events, beware
            of security issues, see : https://lxml.de/api/lxml.etree.XMLParser-class.html
 
-        The niquests library will honor a .netrc-file, if such a file exists
-        username and password may be omitted.  Known bug: .netrc is honored
-        even if a username/password is given, ref https://github.com/python-caldav/caldav/issues/206
+        The requests library will honor a .netrc-file, if such a file exists
+        username and password may be omitted.  Known bugs:
```
tobixen, line 475: Uff ... I made a pull request here some few days ago doing changes in this docstring as well, so it may be needed to do some rebasing here.  I procrastinated merging it, giving the issue author the change to come with comments or do a review ... but  I'll push it to master now :-)

caldav/davclient.py
```diff
@@ -826,6 +833,29 @@ def request(
...
             log.debug("server responded with %i %s" % (r.status_code, r.reason))
+            if (
+                r.status_code == 401
+                and "text/html" in self.headers.get("Content-Type", "")
+                and not self.auth
+            ):
+                # The server can return HTML on 401 sometimes (ie. it's behind a proxy)
+                # The user can avoid logging errors by setting the authentication type by themselves.
+                msg = (
+                    "No authentication object was provided. "
```
tobixen, line 844: I will write some general comments on this shortly

tobixen:
I do have a goal that the caldav library should be as easy as possible to use - both for people that want quick access to their calendar server and for people that don't know any details on the request/niquest libraries, caldav standards, etc.  From that perspective it seems a bit daunting to get warnings telling them to manually set an auth object - so my knee-jerk reaction to this was "don't solve this by adding even more warnings- rather mute the logging if a html page is to be expected".

Anyway I do see the point of avoiding double requests, and those warnings will probably just be ignored by people who aren't into those things, so I will accept this pull request (after it has been rebased with the latest master ... sorry for that one).

My other objection is that I want to move the "best current practice" to use a `caldav.davclient.get_davclient` function rather than constructing a `DAVClient` object from 2.0 on.  From that perspective it may be wrong to ask the user to construct a DAVClient object.  Then again, that's not something you should need to think of, it's my problem, I should fix up `get_davclient` to accept both an auth object, but also the authentication method - and then edit that warning.  I'll create a new issue for that :-) 

 (post-edit rethinking: the `get_davclient` will already accept `auth`, as it passes all unknown connection parameters to the `DAVClient`-constructor.  The `DAVClient` should accept an `auth_type` text string.  This is the way to do it - because a text string can (hopefully) be added to the configuration by some end user using some application (like HomeAssistant), while an end user that doesn't even know Python cannot construct an auth object and send it to DAVClient.  But anyway, that's outside the scope for this pull request - I will take care of this at some point in the future)

tobixen:
Version 2.0 will be released shortly.  Will you complete this one, or should I just fix this myself, or perhaps procrastinate it until 2.1?

tobixen:
I'll try to get done with #523

tobixen:
I did the rebase and merged together the docstring changes.  I'll also fix the error message now to reflect the new `auth_type` connection parameter.

tobixen:
thanks

PR #543, opened 2025-08-19
Use dedicated logger when fixing events

thomasloven:
This change makes the event fixer use the "caldav" logger used otherwise througout the project instead of global logging.

This will allow for complete muting of the warning messages when an event is automatically adjusted.
It's probably easier than forcing e.g. Apple to be standards compliant.

https://github.com/python-caldav/caldav/issues/426#issuecomment-2397075495
https://github.com/home-assistant/core/issues/111633


tobixen:
Thanks

PR #554, opened 2025-09-28
Import missing re module

thyssentishman:
Title says it all.  https://github.com/python-caldav/caldav/blob/55dc53e900bd59cbd38121f7e86a64fa27b75a59/caldav/calendarobjectresource.py#L733


tobixen:
Amazing.  This also means that the line is missing test coverage.  Thanks a lot.

(There exists tools for figuring out missing imports as well as extra imports, I probably ought to do some research on that)

PR #557, opened 2025-11-06
Issue544

tobixen:
More work could probably be done here, but my plate is overfilling, so I'll just merge this to master now


tobixen:
0afd79f and d91fd01 is unrelated.  Should make a separate merge request for those two ..

Issue #102, opened 2020-06-07
Support for rfc6764 - find CalDAV URL through DNS lookup


tobixen:
It may be relevant to support RFC6764.

tobixen:
It took me five and a half years to get time to look into this, and it took Claude some few minutes to churn out an implementation that looks good to me.  Hm.  I'm concerned that we'll be very unemployed all of us quite soon.

(Well, I expect to still spend some days fighting with getting the functional tests pass at as many server implementations and systems as possible)

tobixen:
I hope to merge this feature to the main branch this week, and release version 2.2 this year.

It seems like RFC6764 is much more wide-spread than I thought it would be, so this should definitively have been put in place at least four years ago.

Then again, I do see a major security flaw with the RFC6764 discovery.  If the DNS is not to be trusted, someone can highjack the connection by spoofing the service records, and also spoofing the TLS setting, encouraging the client to connect over plain-text HTTP without certificate validation.  Utilizing this it may be possible to steal the credentials.  This flaw can be mitigated by using DNSSEC, but DNSSEC is not widely used, and there is currently no mechanisms in this package to verify that the DNS is secure.

The AI suggested to add a `require_tls` parameter that defaults to True, it will at least make it a bit more secure.

PR #570, opened 2025-11-27
Add RFC 6764 DNS-based service discovery for CalDAV/CardDAV

tobixen:
Fix for https://github.com/python-caldav/caldav/issues/102

Implements automatic service discovery allowing users to provide just a domain name or email address instead of full URLs. The implementation follows RFC 6764 specification for locating CalDAV and CardDAV services.

Features:
- DNS SRV record lookup (`_caldavs._tcp` / `_caldav._tcp`)
- DNS TXT record lookup for service path information
- Well-known URI fallback (`/.well-known/caldav`)
- Automatic TLS preference with fallback to non-TLS
- Graceful fallback to feature hints if discovery fails
- New caldav.discovery module with comprehensive API
- New enable_rfc6764 parameter for DAVClient (default: True)

Usage examples:
```
  DAVClient(url='user@example.com', ...)        # Email-based discovery
  DAVClient(url='calendar.example.com', ...)    # Domain-based discovery
  DAVClient(url='example.com', enable_rfc6764=False)  # Disable discovery
```

Changes:
- Added `caldav/discovery.py` with complete RFC 6764 implementation
- Enhanced `caldav/davclient.py _auto_url()` with discovery integration
- Added enable_rfc6764 parameter to `DAVClient.__init__()`
- Added dnspython as required dependency in pyproject.toml
- Added example_rfc6764_usage.py demonstrating all usage patterns

Backward compatibility: Fully maintained. Discovery only activates for bare domains/emails. Existing code with full URLs or feature hints works unchanged.


tobixen:
My tasks here:

* [x] Make a new script that will run through all entries in my `tests/conf_private.py` and check if any of them supports the RFC.
* [x]  The said script can probably be moved to the examples folder.
* [x]  Functional testing through the current automated test framework does not work out, as the test servers are locally spun up on localhost through the test framework - but it's possible to make a functional test that does discovery towards some well-known caldav provider, without logging in or utilizing any password
* [x]  Current functional tests towards the servers in `tests/conf_private.py`  should automatically utilize this method, so I need to run through as many servers as possible again.
* [x]  The AI left some noise behind in the project root.  The example file can be moved to the example folder, thoug it may be difficult to set up automated testing on it.  The attempt of describing the changeset in a markdown file should most likely be deleted.

Issue #571, opened 2025-11-27
DNSSEC validation for automatic service discovery


tobixen:
As a continuation of #102 it may be a nice additional feature to validate that the DNS lookups from the service discoveries can be trusted.

PR #572, opened 2025-11-27
Stalled: DNSSEC support

tobixen:
UPDATE: this is stalled at least until 2027.  Currently it does not seem like a good idea.  Support for DNSSEC is too patchy.  Niquests claims to support it, but when looking more into it, the support is rather conditional.

Ref https://github.com/python-caldav/caldav/issues/571

### DIFF
--- a/RFC6764_IMPLEMENTATION.md
+++ b/RFC6764_IMPLEMENTATION.md
@@ -291,20 +291,31 @@ DNSSEC validation adds minimal overhead:
 - No impact on subsequent CalDAV operations
 - Results can be cached to amortize cost
 
-### Limitations
+### How DNSSEC Works in This Implementation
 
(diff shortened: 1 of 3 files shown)


tobixen:
I'm not much happy with this

* In the last commit, we went from validating DNSSEC on the client side to relying on CloudFlare to do the job for us?  Maybe I've misunderstood something, but this does not sound like a change I would like to approve.
* Perhaps the right thing is to work more towards the niquest library and check the library can be improved to do DNSSEC validations in an easier way

Everything considered, I think I will probably not include DNSSEC support in the next release.

tobixen:
DNSSEC may be considered for some future version after 3.0.  In the very best case, 2026-Q2, but possibly never.

PR #573, opened 2025-11-27
Add Baikal Docker test server framework for CI/CD

tobixen:
Set up infrastructure for running tests against Baikal CalDAV server in Docker containers, both locally and in GitHub Actions.

Changes:
- Added Baikal service container to GitHub Actions workflow
- Created docker-compose setup under tests/docker-test-servers/baikal/
- Added Baikal configuration module (tests/conf_baikal.py)
- Created comprehensive documentation and helper scripts
- Updated .gitignore to exclude backup directories
- Added tests/README.md with testing guide

The GitHub Actions workflow now automatically spins up a Baikal container and makes it available at http://localhost:8800 for testing.


cbcoutinho:
Hey @tobixen, this is a great addition - I've been using a similar pattern in https://github.com/cbcoutinho/nextcloud-mcp-server for integration testing.

Would you be willing to add Nextcloud or Radicale as well for multi-CalDav server testing? I haven't added my docker-compose to #565 to keep the PR focused on `httpx`, but it's definitely useful - albeit results in longer test times with all the cicd setup

tobixen:
Yes, if I can get baikal working, then I will be looking into adding other docker-based calendar servers.

Radicale is not on the list.  Radicale and Xandikos is already included in the standard test suite, since they are python-based I was able to launch an instance from the test code without involving docker, so the tests are already running towards radicale regularly.

tobixen:
This was a hard nut to crack, I'm happy that I could outsource it to the AI.  It was having a really hard time getting the baikal server set up, oscillating between trying to mount some directory pre-seeded with configuration versus copying the configuration into the server after startup.  All the time baikal just wanted to redirect to an install script.

PR #578, opened 2025-11-29
Docker server testing cyrus

tobixen:
Follows the same framework as baikal and nextcloud


tobixen:
This changeset contains quite a lot more than just the bare minimum for spinning up the Cyrus container.  I had to do lots of work to get the tests passing.  I tried doing some git rebasing to make different things in different commits, but it was too difficult.  The changeset includes:

* quite some fixing of test code, and also some few new tests
* improvements to the compatibility hints system (particularly moving stuff from the old "incompatibility_flags" structure to the new "feature set" structure (and with tests made in parallell to the caldav-server-tester project).
* calendar.delete() now has logic for removing everything on the calendar, if the server does not support delete.
* principal_search will now raise an error if the server delivers anything higher than a 2xx-reply
* bugfix in search - the `post_filter`-boolean was not propagated down when doing a multi-component search towards a server requiring a component type to be given.
* Containers should not leave the ephemeral volume behind after tests are run

Issue #587, opened 2025-12-04
Calendar.search broken with TypeError: Calendar.search() got multiple values for argument 'sort_keys'


simon-ging:
Hi,

In version 2.1.2 I get an error 

```python
        events_fetched = main_cal.search(
            start=now - timedelta(days=1, hours=2), end=now - timedelta(days=1)
        )
```

```
    events_fetched = main_cal.search(                                                                     
                     ^^^^^^^^^^^^^^^^             
  File ".../lib/python3.12/site-packages/caldav/collection
.py", line 1012, in search             
    return self.search(                                                                                   
           ^^^^^^^^^^^^                                                                                   
TypeError: Calendar.search() got multiple values for argument 'sort_keys'    
```

Downgrading to 2.0.1 fixes it

The search method code is a forest of try except if else, so it's not possible for me to see what's happening at first glance

Also I cannot create a reproducible example because it's my private calendar

Hope this still helps someone,

Best,

Simon

tobixen:
search.py is not part of the v2.1.2-release, so this error is probably in the master branch.  My plan was to release it today, so this issue has top priority.

I'm not much happy with how search.py turned out - I already made a note in #580 that it's needed to refactor it.  Nevertheless, I'm a bit in a rush to get 2.2.0 released.

Do you have the full stack trace?

tobixen:
Also, what server do you use?

tobixen:
Please disregard my earlier message (now deleted).  I misread the issue description a bit.  (but the full stack-trace is still valuable for me)

It would be very helpful if you could check if the problem persists in the current master branch.

I asked Claude for help, and it identified a bug in commit 507b080d.  I'm investigating further.

tobixen:
The buggy code has been refactored and is not present in the current master branch.  Please retry with the master branch if possible.

tobixen:
I will close this issue.

* Assumed fixed in master branch, which (I hope) will be released later today as 2.2.0
* Assumed fixed in pull request #588 which (I hope) will be released as 2.1.3 later today

Please comment if the issue persists in master, 2.2.0 or 2.1.3.

tobixen:
The fix proposed by Claude broke the tests, so there will be no 2.1.3.  However, 2.2 should not have this issue, and it was released some minutes ago.

PR #588, opened 2025-12-04
Fix duplicate parameter bug in search() recursive call

tobixen:
This pull request is based on the v2.1.2-release and could possibly go in as a 2.1.3-release.

Fixed issue https://github.com/python-caldav/caldav/issues/587 where the recursive search() call had multiple bugs:
1. Used invalid syntax *kwargs2 instead of **kwargs2 to unpack dict
2. Passed sort_reverse both explicitly and in kwargs2 dict

This caused TypeError: Calendar.search() got multiple values for argument 'sort_keys' when the backward compatibility fallback code path was triggered (when a server returns an error on searches without explicit comp_class).

The bug was introduced in commit 507b080d during compatibility hints refactoring and wasn't caught earlier because this code path is only executed for specific server error conditions.

### DIFF
--- a/caldav/collection.py
+++ b/caldav/collection.py
@@ -976,7 +976,6 @@ class Calendar(DAVObject):
             ## those arguments:
             kwargs2 = {
                 "include_completed": include_completed,
-                "sort_reverse": sort_reverse,
                 "expand": expand,
                 "server_expand": server_expand,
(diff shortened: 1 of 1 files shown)


tobixen:
Tests failed, so I will abandon the idea of doing a  v2.1.2 maintenance release.

PR #603, opened 2025-12-09
Playground/new async api design

tobixen:
I'm playing a bit with Claude and asking it for advises on the async API.

This in parallell with #565 

Currently discussing how the davclient API may be improved in an async version


tobixen:
This is not a playground anymore, this has become a candidate for the v3.0-release.  I will close this pull request and make a new rebased pull request.

PR #604, opened 2025-12-09
fix/examples use get davclient

tobixen:
Update examples to use get_davclient() instead of DAVClient()


tobixen:
Wrong branch possibly

PR #615, opened 2026-01-22
FIx for issue 614

tobixen:
... missing import.  I think probably ruff would catch this.  It's on my list to introduce ruff.


tobixen:
It seems like I had some more minor bugfixes here that wasn't pushed earlier

Issue #647, opened 2026-03-21
Calling `c.events()` seems to break my python script sometimes


ShawkMusic:
When I run a python script that uses caldav (more details below), the script freezes for like a minute or two and then I get this error:

```
Traceback (most recent call last):
  File "/nix/store/n3a01p8l886s3v5n4x7frg52y3hppnfs-python3-3.13.12-env/lib/python3.13/site-packages/urllib3_future/response.py", line 668, in _error_catcher
    yield
  File "/nix/store/n3a01p8l886s3v5n4x7frg52y3hppnfs-python3-3.13.12-env/lib/python3.13/site-packages/urllib3_future/response.py", line 771, in _raw_read
    data = self._fp_read(amt) if not fp_closed else b""
           ~~~~~~~~~~~~~^^^^^
  File "/nix/store/n3a01p8l886s3v5n4x7frg52y3hppnfs-python3-3.13.12-env/lib/python3.13/site-packages/urllib3_future/response.py", line 756, in _fp_read
    return self._fp.read(amt) if amt is not None else self._fp.read()
                                                      ~~~~~~~~~~~~~^^
  File "/nix/store/n3a01p8l886s3v5n4x7frg52y3hppnfs-python3-3.13.12-env/lib/python3.13/site-packages/urllib3_future/backend/_base.py", line 342, in read
    data, self._eot, self.trailers = self.__internal_read_st(
                                     ~~~~~~~~~~~~~~~~~~~~~~~^
        __size, self._stream_id
        ^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/nix/store/n3a01p8l886s3v5n4x7frg52y3hppnfs-python3-3.13.12-env/lib/python3.13/site-packages/urllib3_future/backend/hface.py", line 1544, in __read_st
    events: list[DataReceived | HeadersReceived] = self.__exchange_until(  # type: ignore[assignment]
                                                   ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        DataReceived,
        ^^^^^^^^^^^^^
    ...<5 lines>...
        respect_end_stream_signal=__respect_end_signal,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

Issue #620, opened 2026-02-03
sleep-retry-logic


tobixen:
Some servers sometimes have transient errors.

My idea is to make it possible for the client to sleep and retry on transient errors.  This should be a configurable opt-in feature.

It goes in under the "features"-section in the server configuration.  Perhaps this "feature" can be called "high-availability" with type "server-peculiarity".  Or perhaps it should be more explicit,  "retry-on-failure" of type "client-feature".  Extra configuration parameters should indicate the sleep before the first retry (one may want to keep it short - perhaps all that is needed is a retry, without any sleep), the sleep between subsequent retries, the max number of retries.  May consider supporting growing sleep intervals too.

tobixen:
(The text below was added by Claude, the AI.  I think PUT and DELETE should be idempotent in CalDAV, and POST is not used that often.   Exponential back-off may not make sense in cases like network outages)

## Design Suggestions

### Naming
I'd suggest `retry-on-transient-error` over `high-availability` - the latter might be misleading since this handles server instability rather than true HA (load balancing, failover, etc.).

### Configuration Structure
```yaml
retry-on-transient-error:
  enabled: true
  max_retries: 3
  initial_delay: 0.5        # seconds before first retry
  backoff_factor: 2.0       # exponential backoff multiplier
  max_delay: 30             # cap on delay between retries
  jitter: true              # add randomness to avoid thundering herd
```

### Which Errors to Retry
Consider categorizing errors:

**Transient (retry by default):**
- HTTP 500, 502, 503, 504 - server errors
- HTTP 429 - rate limited (should respect Retry-After header if present)
- Connection timeouts/resets

**Context-dependent:**
- AuthorizationError (403) - usually permanent, but robur shows it can be transient for newly created resources

**Permanent (never retry):**
- HTTP 400, 401, 404 - client errors
- HTTP 405, 409 - method/conflict errors

### Idempotency Consideration
Safe to retry: GET, HEAD, PROPFIND, REPORT, OPTIONS
Risky to retry: PUT, POST, DELETE, MKCALENDAR (might succeed but return error)

Could add `retry_safe_methods_only: true` option to only retry read operations.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

tobixen:
This issue got duplicated, we'll continue tracking it in #695.  See also design document `docs/design/RETRY_AND_RESILIENCE_DESIGN.md`.  Comments from Claude Opus below.

---

[...]

PR #648, opened 2026-03-21
stalled: catch and retry on ConnectionError/Timeout from niquests (issue #647)

tobixen:
UPDATE: I'm leaving this pull request to rot for a while as I'm not 100% confident it's pulling the project in the right direction.

---

This commit was done by Claude and hasn't been thoroughly reviewed yet.  The code should be safe to run, but I'm wondering if this is the right way of fixing the problem.  Pull request is in "draft mode" as for now.

---

niquests uses lazy HTTP/2 response gathering: accessing r.status_code or
r.headers can raise ConnectionError if the underlying connection has gone
stale.  This produces a confusing nested-exception traceback and an
intermittent failure for callers (c.events(), etc.).

Changes:
- Add DAVNetworkError(DAVError) to caldav/lib/error.py to represent
  network-level failures wrapping the underlying library exception.
- In DAVClient._sync_request(), wrap the session.request() call and all
  subsequent response-attribute access in a try/except that catches
  requests.exceptions.ConnectionError and requests.exceptions.Timeout,
  re-raising as DAVNetworkError.
- In DAVClient.request(), catch DAVNetworkError and retry once for
  idempotent methods (GET, HEAD, OPTIONS, PROPFIND, REPORT, PUT, DELETE,
  MKCOL, MKCALENDAR).  POST is the only non-idempotent method and is not
  retried.
- Apply the same pattern to AsyncDAVClient._async_request() using an
  outer try/except around the existing auth-workaround try/except, with
  a _connection_retried flag to prevent infinite loops.

The retry transparently recovers from stale HTTP/2 connection reuse, which
is the root cause of the "sometimes works, sometimes times out" symptom.

Ref: https://github.com/python-caldav/caldav/issues/647

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/caldav/async_davclient.py
+++ b/caldav/async_davclient.py
@@ -69,6 +69,7 @@ from caldav import __version__
 from caldav.base_client import BaseDAVClient
 from caldav.base_client import get_davclient as _base_get_davclient
 from caldav.compatibility_hints import FeatureSet
+from caldav.davclient import _SAFE_METHODS
 from caldav.lib import error
 from caldav.lib.python_utilities import to_wire
 from caldav.lib.url import URL
@@ -91,6 +92,12 @@ from caldav.requests import HTTPBearerAuth
 from caldav.response import BaseDAVResponse
 
 log = logging.getLogger("caldav")
+
+# Network-level exceptions for the active async HTTP library.
+if _USE_HTTPX:
+    _NETWORK_EXCEPTIONS: tuple = (httpx.ConnectError, httpx.TimeoutException, httpx.NetworkError)
+else:
+    _NETWORK_EXCEPTIONS = (niquests.exceptions.ConnectionError, niquests.exceptions.Timeout)
 
 if sys.version_info < (3, 11):
     from typing_extensions import Self
@@ -404,6 +411,7 @@ class AsyncDAVClient(BaseDAVClient):
         method: str = "GET",
         body: str = "",
         headers: Mapping[str, str] | None = None,
+        _connection_retried: bool = False,
     ) -> AsyncDAVResponse:
         """
         Async HTTP request implementation with auth negotiation.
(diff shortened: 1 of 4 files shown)


tobixen:
⚠️ Most of this comment is AI-generated (Claude Sonnet 4.6 via Claude Code) on behalf of tobixen

## Review considerations

### Root cause uncertainty

The linked niquests issue ([jawah/niquests#352](https://github.com/jawah/niquests/issues/352)) was closed as "Not Planned" with the maintainer saying the Claude analysis was hallucinated and the actual problem might be HTTP/3 (Alt-Svc upgrade) rather than HTTP/2 stale connections. The fix might not actually resolve the reporter's issue, and if the real root cause is something else (e.g., silent Alt-Svc protocol upgrade), the retry might just add a second full-timeout wait before failing.

That said, catching transient network errors and retrying idempotent requests is a standard HTTP client pattern — good defense-in-depth regardless of root cause.

### Bug: DELETE retry can raise confusing NotFoundError

A silly human is insisting that it's relatively safe to repeat both `DELETE` and `PUT`, only `POST` (not even used by the CalDAV protocol) may cause nasty side-effects like double-transactions.  However, it should be considered a bug that `DELETE` is in `_SAFE_METHODS`. If the first DELETE succeeds silently (server processed it but the response was lost), the retry will get a 404, which `DAVResponse` raises as `NotFoundError`. From a user's perspective, getting `NotFoundError` from `event.delete()` implies the event wasn't there in the first place — not that the delete succeeded. Consider removing DELETE from `_SAFE_METHODS`, or catching `NotFoundError` on retry and converting it to success.

### Async inner `except Exception` interaction

The async refactor wraps everything in an outer `try/except _NETWORK_EXCEPTIONS`, but the inner auth-workaround still uses `except Exception`. When a client is configured with `password=` but no `auth=` (common case), network errors during the initial request hit the inner `except Exception` first — which then tries auth detection instead of being caught by the outer `_NETWORK_EXCEPTIONS` handler (the condition is `if self.auth or not self.password: raise` — if you have a password but no auth object yet, network errors silently trigger auth-detection instead of the retry path).

This is a pre-existing issue, not introduced by this PR, but the restructuring makes it harder to reason about. The outer retry logic won't fire in this common case for async clients.

### `DAVNetworkError` not exported at top level

Consistent with other errors (`NotFoundError`, `AuthorizationError`), but since this is a new user-catchable exception, worth documenting in release notes.

### No async unit tests

The 6 new unit tests only cover the sync `DAVClient`. The async retry path (`_async_request` with `_connection_retried`) has no coverage, and the async restructuring is the riskier change.

### CHANGELOG not updated

The fix for issue #647 isn't in the 3.1.x changelog.

### Timeout doubles on retry

If the original failure was a timeout after 2 minutes, the retry may also time out after 2 minutes. The fix improves eventual recovery but not latency. Users with `timeout=None` can still see very long hangs. The advice to set an explicit timeout remains relevant.

### What's good

- `_SAFE_METHODS` frozenset is otherwise correct
- `except error.DAVError: raise` before `except _NETWORK_EXCEPTIONS` correctly prevents swallowing auth/rate-limit errors
- The `_connection_retried` flag correctly prevents infinite loops
- `raise error.DAVNetworkError(...) from e` preserves the original traceback chain
- Tests cover the key scenarios (REPORT retry, POST no-retry, double-failure raises)

### Suggested actions before merging

1. Remove DELETE from `_SAFE_METHODS` (or handle 404 on retry gracefully)
2. Add async unit tests for the retry path
3. Update CHANGELOG
4. Ask the reporter to test the branch — we still don't know if the fix actually resolves their issue

⚠️ Most of this comment is AI-generated (Claude Sonnet 4.6 via Claude Code) on behalf of tobixen

tobixen:
This is somewhat related to #620

tobixen:
Closing this pull request, the design has been reconsidered.  Retries belongs to the http-library, not to the caldav library - and it should also be consolidated with the rate-limit wait-and-retry-logic.  Claude has spewed out some big blob of text, see below, it's TLDR for me, so I won't guarantee it's correct.

---

⚠️ AI-generated by Claude Opus 5 via Claude Code on behalf of tobixen, from the prompt(s): "I just noticed that issue #695, pull request #648 and issue #620 is related" / "Write up the design doc.  Close two of the three issues as duplicate, fix the non-closed issue to point to the design document.  Close the PR if the design document differs too much.  Do some research on what batteries are included in the different http libraries (httpx2, niquests)"

Closing this.  The hunch in the UPDATE at the top - "not 100% confident it's pulling the project in the right direction" - turned out to be correct, and it is now written down: `docs/design/RETRY_AND_RESILIENCE_DESIGN.md` on `master`, section 5.  The short version is that this pull request is not wrong about the bug, it is at the wrong layer.

Retries for the two transport failure classes - "the request never reached the server" and "the response never came back" - belong in the HTTP library, which already implements them properly.  `niquests.Session` and `AsyncSession` take `retries=` straight in the constructor (`niquests.RetryConfiguration` is a re-export of urllib3-future's `Retry`), and `requests` takes the same object via a mounted `HTTPAdapter`.  That gets us a retry budget, backoff with jitter, `Retry-After` handling and a method allow-list, none of which this pull request has, and it also covers failures that happen *inside* the connection pool where `DAVClient.request()` cannot see them at all.  Notably, the 105-line async restructuring here - the part the review flagged as the riskiest - collapses to a single constructor argument.

There is one thing no amount of `Retry` configuration can fix, and this pull request is right about it: niquests gathers HTTP/2 responses lazily, so `r.status_code` can raise `ConnectionError` *after* `adapter.send()` has returned, outside urllib3's retry scope.  The traceback in https://github.com/python-caldav/caldav/issues/647 shows it firing from the `log.debug("server responded with %i %s" % (r.status_code, r.reason))` line in `request()`.  That needs caldav-side code - roughly 15 lines per client rather than 250 across four files - and, together with `DAVNetworkError`, it is what should be resubmitted as a small pull request.  `DAVNetworkError` earns its place on its own: without it a caller has to catch `niquests.exceptions.ConnectionError` *or* `requests.exceptions.ConnectionError` *or* `httpx.TransportError` depending on which library happens to be installed, which rather defeats the point of having a fallback chain.

Two corrections to my own earlier review comment on this pull request:

* **The DELETE objection is withdrawn.**  `DAVObject._post_delete()` already accepts 404, and `DAVResponse` does not raise on a 404 for a plain DELETE, so a retried DELETE whose first attempt silently succeeded does not surface as `NotFoundError`.  `DELETE` and `PUT` can stay retryable - the silly human was right, and so is urllib3, whose default `allowed_methods` includes both.
* **A plain `Retry(3)` would not have fixed the reported symptom**, which is a subtlety neither the review nor the pull request noticed: urllib3 classifies "Remote end closed connection without response" as a `ProtocolError`, therefore as a *read* error, and read errors are only retried for methods in `allowed_methods` - whose default is `{OPTIONS, TRACE, DELETE, GET, HEAD, PUT}`, containing neither `PROPFIND` nor `REPORT`.  The allow-list has to be set explicitly, which is another argument for configuring this at session construction rather than per request.

Further work is tracked in https://github.com/python-caldav/caldav/issues/695.

⚠️ AI-generated by Claude Opus 5 via Claude Code on behalf of tobixen.

PR #688, opened 2026-06-19
fix(jmap): correctness fixes + dedup from June 2026 review

tobixen:
I managed to run a full code review of the caldav library with the Claude Fable model before it was yanked, and I've had Claude Sonnet and Claude Opus helping me fixing the things.  I'm still busy reviewing, rewording, squashing similar commits together, etc.

The experimental JMAP support was written by @SashankBhamidi - perhaps you could help me reviewing this particular changeset?  (or rewrite it completely if you prefer that).

The rest of this message is AI-generated

---

## JMAP fixes from the June 2026 code review

This consolidates the JMAP-related findings from the full codebase review
(`docs/design/FULL_CODE_REVIEW_2026-06.md`) into a single commit, so the JMAP
backend can be fixed on `master` independently of the larger v3.3.0 work.

### Correctness fixes
- **§1.3** `create_task()` was missing the empty-`created`-dict guard that
  `create_event()` already had → bare `KeyError` instead of `JMAPMethodError`
  (sync + async).
- **§4.1** `jscal_to_ical`: override child VEVENT with no `start` patch key took
  DTSTART from the master instead of the override key — relocated every
  title-only override to the master's first occurrence.
- **§4.2** `jscal_to_ical`: EXDATE / RECURRENCE-ID were always emitted as naive
  floating DATE-TIMEs; a floating EXDATE on a TZID event matches no instance, so
  excluded occurrences reappeared. Now parsed with the event's value type.
- **§4.3** `_format_local_dt()`: stripped the `Z` suffix — RFC 8984 LocalDateTime
  (override keys, `until`) must be suffix-less or strict servers won't match.
- **§4.4** STATUS was silently dropped both conversion directions
  (`CANCELLED` round-tripped as `confirmed`). Added the CONFIRMED/TENTATIVE/
  CANCELLED mappings.
- **§4.5** `update_event` sent the full object as the PatchObject, so removed
  properties (LOCATION, VALARM, …) silently persisted. Absent optional props are
  now set to `null` (RFC 8620 §3.3 semantics).
- **§4.6** `search()` passed datetimes through `isoformat()` instead of the
  `...Z` UTCDate format JMAP requires.
- **§4.7** `get_objects_by_sync_token()` discarded `newState`, forcing a racey
  follow-up `get_sync_token()`. Now returns `(added, modified, deleted, new_sync_token)`.

### Dedup / perf
- **§5.1** Response-parsing logic moved into shared static methods on
  `_JMAPClientBase`; sync/async public methods are now thin wrappers
  (`async_client.py` 550 → 415 lines), so future fixes live in one place.
- **§5.5** One persistent HTTP session per client instead of per request.

### Testing
`tests/test_jmap_unit.py` — 266 passed.

### DIFF
--- a/caldav/jmap/convert/_patch.py
+++ b/caldav/jmap/convert/_patch.py
@@ -1,0 +1,33 @@
+"""
+RFC 8620 PatchObject helpers for CalendarEvent/set update calls.
+
+When updating an event, absent keys preserve the server's current value.
+To delete an optional property the patch must set it to null explicitly.
+"""
+
+from __future__ import annotations
+
+# Optional JSCalendar top-level properties that must be explicitly nulled in
+# a CalendarEvent/set update when they are absent from the converted result.
+# This ensures properties removed client-side (e.g. LOCATION deleted from
+# the iCalendar) are actually removed on the server, not silently preserved.
+_NULL_FOR_UPDATE: frozenset[str] = frozenset(
+    {
+        "description",
+        "color",
+        "locations",
+        "keywords",
+        "priority",
+        "privacy",
+        "freeBusyStatus",
+        "status",
+        "sequence",
+        "showWithoutTime",
+        "timeZone",
+        "recurrenceRules",
+        "excludedRecurrenceRules",
+        "recurrenceOverrides",
+        "participants",
+        "alerts",
+    }
+)
--- a/caldav/jmap/convert/_utils.py
+++ b/caldav/jmap/convert/_utils.py
@@ -111,11 +111,12 @@ def _duration_to_timedelta(duration_str: str) -> timedelta:
 
 
 def _format_local_dt(dt: datetime | date) -> str:
-    """Format a datetime or date as a JSCalendar LocalDateTime or UTCDateTime string.
+    """Format a datetime or date as a JSCalendar LocalDateTime string.
 
-    JSCalendar uses:
-      - LocalDateTime: "2024-03-15T09:00:00"    (no TZ suffix)
-      - UTCDateTime:   "2024-03-15T09:00:00Z"   (uppercase Z)
+    RFC 8984 requires LocalDateTime (no Z suffix) for override keys and RRULE
+    ``until`` values.  Timezone information is stripped — callers must convert
+    UTC datetimes to the event's local timezone before calling if the event uses
+    TZID; for floating or all-day events the naive value is already correct.
 
     For date objects (all-day), uses T00:00:00 suffix.
 
@@ -123,10 +124,8 @@ def _format_local_dt(dt: datetime | date) -> str:
         dt: A datetime (with or without tzinfo) or a date.
 
     Returns:
-        Formatted string suitable for use as a JSCalendar override key or datetime value.
+        Formatted string suitable for use as a JSCalendar override key or RRULE until.
     """
     if isinstance(dt, datetime):
-        if dt.tzinfo is not None and dt.utcoffset() == timedelta(0):
-            return dt.strftime("%Y-%m-%dT%H:%M:%SZ")
         return dt.strftime("%Y-%m-%dT%H:%M:%S")
     return f"{dt.isoformat()}T00:00:00"
--- a/caldav/jmap/convert/ical_to_jscal.py
+++ b/caldav/jmap/convert/ical_to_jscal.py
@@ -12,11 +12,39 @@ from __future__ import annotations
 
 import uuid
 from datetime import date, datetime, timedelta
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import icalendar
 
 from caldav.jmap.convert._utils import _format_local_dt, _timedelta_to_duration
 from caldav.lib import vcal
+
+# RFC 5545 STATUS -> RFC 8984 status; module-level constant (cf. _CLASS_MAP etc.)
+_STATUS_ICAL_TO_JSCAL = {
+    "CONFIRMED": "confirmed",
+    "TENTATIVE": "tentative",
+    "CANCELLED": "cancelled",
+}
+
+
+def _to_event_local(dt, time_zone):
+    """Shift a tz-aware datetime to naive wall-clock in the event's timeZone.
+
+    RFC 8984 LocalDateTime values (RRULE ``until``, EXDATE override keys) are
+    expressed in the recurring event's own time zone.  A UTC UNTIL/EXDATE on a
+    TZID event must therefore be converted to that zone before being formatted
+    as a (suffix-less) LocalDateTime, otherwise the series ends at the wrong
+    wall-clock time and the round-trip emits a Z-less UNTIL that RFC 5545
+    §3.3.10 forbids for TZID events.  Floating / all-day / naive values (and
+    unknown time zones) pass through unchanged.
+    """
+    if time_zone and isinstance(dt, datetime) and dt.tzinfo is not None:
+        try:
+            return dt.astimezone(ZoneInfo(time_zone)).replace(tzinfo=None)
+        except ZoneInfoNotFoundError:
+            return dt
+    return dt
+
 
 _CLASS_MAP = {
     "PRIVATE": "private",
@@ -68,11 +96,14 @@ def _dtstart_to_jscal(dtstart_prop) -> tuple[str, str | None, bool]:
     return dt.strftime("%Y-%m-%dT%H:%M:%S"), None, False
 
 
-def _rrule_to_jscal(rrule_prop) -> dict:
+def _rrule_to_jscal(rrule_prop, time_zone=None) -> dict:
     """Convert an iCalendar RRULE property to a JSCalendar RecurrenceRule dict.
 
     Always emits @type, interval, rscale, skip, firstDayOfWeek to match the
     fields Cyrus returns — makes round-trip comparison predictable.
+
+    ``time_zone`` is the event's IANA time zone; a UTC ``UNTIL`` is shifted to
+    it so the LocalDateTime value matches the event frame (see _to_event_local).
     """
     rule: dict = {
         "@type": "RecurrenceRule",
@@ -97,7 +128,7 @@ def _rrule_to_jscal(rrule_prop) -> dict:
 
     until_list = rrule_prop.get("UNTIL", [])
     if until_list:
-        rule["until"] = _format_local_dt(until_list[0])
+        rule["until"] = _format_local_dt(_to_event_local(until_list[0], time_zone))
 
     byday_list = rrule_prop.get("BYDAY", [])
     if byday_list:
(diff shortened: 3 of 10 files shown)


SashankBhamidi:
Hi, is it okay if I review this on Sunday or later? I'm currently drowned in a lot of business work right now.

tobixen:
Sure.  While I had hoped to release caldav 3.3.0 this week, realistically I will not manage.

caldav/jmap/convert/_utils.py
```diff
@@ -111,22 +111,21 @@ def _duration_to_timedelta(duration_str: str) -> timedelta:
 
 
 def _format_local_dt(dt: datetime | date) -> str:
```
SashankBhamidi, line 113: The docstring now says "callers must convert UTC datetimes to the event's local timezone before calling if the event uses TZID" but none of the three callers in `ical_to_jscal.py` actually do that (at lines 100, 163, and 312 of that file). I ran it to confirm the data bug:

```python
# TZID=Europe/Berlin event, UNTIL=20240701T120000Z in the RRULE
jscal = ical_to_jscal(ical)
jscal["recurrenceRules"][0]["until"]
# => "2024-07-01T12:00:00"   (UTC wall-clock, wrong)
# should be "2024-07-01T14:00:00"  (Berlin local, UTC+2)
```

The round-trip back through `jscal_to_ical` produces `UNTIL=20240701T120000` with no Z suffix, which RFC 5545 §3.3.10 forbids for TZID events. The recurrence series ends two hours early for UTC+2 users. The same applies at `ical_to_jscal.py:163` when an EXDATE is expressed in UTC on a TZID event, though that form is less common since most clients emit EXDATE with the TZID parameter matching DTSTART.

The fix is to thread the event timezone into `_rrule_to_jscal` and `_exdate_to_overrides` so they can call `dt.astimezone(ZoneInfo(tz)).replace(tzinfo=None)` before `_format_local_dt`. The timezone is already extracted at `ical_to_jscal.py:64` from `dtstart_prop.params.get("TZID")` but is not passed to either helper. I would keep this in the same PR rather than deferring it since the UNTIL case is silent data corruption on a common event pattern.

caldav/jmap/convert/jscal_to_ical.py
```diff
@@ -353,6 +353,17 @@ def jscal_to_ical(jscal: dict) -> str:
...
+    status = jscal.get("status")
+    if status:
+        _STATUS_JSCAL_TO_ICAL = {
+            "confirmed": "CONFIRMED",
+            "tentative": "TENTATIVE",
+            "cancelled": "CANCELLED",
+        }
+        ical_status = _STATUS_JSCAL_TO_ICAL.get(status)
+        if ical_status:
+            event.add("status", ical_status)
```
SashankBhamidi, line 382: The STATUS mapping dict is defined inside the `if status:` block on every call. Every other mapping in this file (`_PRIVACY_TO_CLASS`, `_FREE_BUSY_TO_TRANSP`, `_PARTSTAT_MAP`, `_KIND_TO_CUTYPE`) is a module-level constant. Non-blocking, but inconsistent with the rest of the file. Hoist `_STATUS_JSCAL_TO_ICAL` to module level and simplify this block to:

```suggestion
    status = jscal.get("status")
    if status:
        ical_status = _STATUS_JSCAL_TO_ICAL.get(status)
        if ical_status:
            event.add("status", ical_status)
```

caldav/jmap/convert/ical_to_jscal.py
```diff
@@ -386,6 +386,17 @@ def ical_to_jscal(ical_str: str, calendar_id: str | None = None) -> dict:
...
+    status = master.get("STATUS")
+    if status:
+        _STATUS_ICAL_TO_JSCAL = {
+            "CONFIRMED": "confirmed",
+            "TENTATIVE": "tentative",
+            "CANCELLED": "cancelled",
+        }
+        jscal_status = _STATUS_ICAL_TO_JSCAL.get(str(status).upper())
+        if jscal_status:
+            jscal["status"] = jscal_status
```
SashankBhamidi, line 427: Same issue: `_STATUS_ICAL_TO_JSCAL` is rebuilt on every call. Should be a module-level constant alongside `_CLASS_MAP`, `_PARTSTAT_MAP`, etc.

```suggestion
    status = master.get("STATUS")
    if status:
        jscal_status = _STATUS_ICAL_TO_JSCAL.get(str(status).upper())
        if jscal_status:
            jscal["status"] = jscal_status
```

tests/test_jmap_unit.py
```diff
@@ -1604,6 +1660,112 @@ def test_update_event_drops_uid_from_patch(self, monkeypatch):
...
+        # {"locations": null, "description": null, ...} so the server removes them.
+        _ICAL_WITH_LOCATION = (
+            "BEGIN:VCALENDAR\r\nVERSION:2.0\r\n"
+            "BEGIN:VEVENT\r\n"
+            "UID:loc-uid@example.com\r\n"
+            "DTSTART:20240615T090000Z\r\n"
+            "SUMMARY:Event with Location\r\n"
+            "LOCATION:Old Conference Room\r\n"
+            "END:VEVENT\r\nEND:VCALENDAR\r\n"
+        )
```
SashankBhamidi: `_ICAL_WITH_LOCATION` is assigned but never passed to anything. Only `_ICAL_WITHOUT_LOCATION` is used at line 1688. Looks like scaffolding from an earlier draft. Non-blocking but should be dropped.

```suggestion
        _ICAL_WITHOUT_LOCATION = (
            "BEGIN:VCALENDAR\r\nVERSION:2.0\r\n"
            "BEGIN:VEVENT\r\n"
            "UID:loc-uid@example.com\r\n"
            "DTSTART:20240615T090000Z\r\n"
            "SUMMARY:Event without Location\r\n"
            "END:VEVENT\r\nEND:VCALENDAR\r\n"
        )
```

SashankBhamidi (changes requested):
Thanks for running the full codebase review and pulling me in on the JMAP sections.

Most of this is correct and the §5.1 refactor is genuinely the right move structurally. Having every public async method be a three-line wrapper over `_request()` and a shared static parser means future fixes no longer need to be made in two places, which is exactly how the §1.3 `create_task` bug got there in the first place. The §4.5 implementation is also more careful than the original finding required: the `_unsupported_null_keys` retry loop only retries when every property in the server rejection is null-cleanup we injected, terminates because each iteration strictly shrinks the patch, and surfaces genuine `invalidProperties` errors without retrying. I traced through it.

One finding needs fixing before merge. The §4.3 change to `_format_local_dt` is correct in what it does but leaves the callers in an inconsistent state for TZID events with RRULE UNTIL clauses. Details in the inline comment. The STATUS maps and the dead test variable are non-blocking.

On the two bot findings: `_NULL_FOR_UPDATE` is not unused. It is imported at `client.py:46` and used at line 178. Dismiss that one. The `_ICAL_WITH_LOCATION` finding is correct, noted below.

SashankBhamidi:
> Sure. While I had hoped to release caldav 3.3.0 this week, realistically I will not manage.

Sorry for the delay! Back to you, good to merge after my comments have been addressed. 

I'll be afk for a few days so feel free to merge.

tobixen:
Well, I'm also delayed, hoped to get v3.3.0 launched yesterday, but it seems like there will be several days until I have time looking more into the caldav library.

caldav/jmap/convert/_patch.py
```diff
@@ -0,0 +1,33 @@
...
+To delete an optional property the patch must set it to null explicitly.
+"""
+
+from __future__ import annotations
+
+# Optional JSCalendar top-level properties that must be explicitly nulled in
+# a CalendarEvent/set update when they are absent from the converted result.
+# This ensures properties removed client-side (e.g. LOCATION deleted from
+# the iCalendar) are actually removed on the server, not silently preserved.
+_NULL_FOR_UPDATE: frozenset[str] = frozenset(
```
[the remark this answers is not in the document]
tobixen: ⚠️ This comment is AI-generated (Claude Opus 4.8 via Claude Code) on behalf of tobixen

False positive: `_NULL_FOR_UPDATE` is used — imported at `caldav/jmap/client.py:46` and iterated at `:178` (the null-clearing loop in `update_event`); not removing it.

⚠️ This comment is AI-generated (Claude Opus 4.8 via Claude Code) on behalf of tobixen

tobixen:
Thanks @SashankBhamidi — I had Claude go through this, I've looked through the changes and it seems like all the points you've raised have been resolved - do you agree?  I've also verified that the new tests fails on 0204af4 but passes with 43226c5.  I'm also asking it to come up with some integration tests.

Below is the AI-generated description of the changes done.

---

⚠️ This comment is AI-generated (Claude Opus 4.8 via Claude Code) on behalf of tobixen

**§4.3 (blocking) — fixed.** Threaded the event TZID into `_rrule_to_jscal` and `_exdate_to_overrides`, so a UTC `UNTIL`/`EXDATE` on a TZID event is now shifted to event-local wall-clock before formatting (your Berlin example now yields `2024-07-01T14:00:00`). I also fixed the jscal→ical direction: `_jscal_rrule_to_rrule` now converts a LocalDateTime `until` back to UTC, so the round-trip emits `UNTIL=…120000Z` with the `Z` suffix that RFC 5545 §3.3.10 requires for a TZID `DTSTART` (it previously emitted a forbidden suffix-less value). Added two regression tests covering the UTC+2 `UNTIL` and `EXDATE` round-trips.

**Nits — done.** Hoisted `_STATUS_ICAL_TO_JSCAL` and `_STATUS_JSCAL_TO_ICAL` to module-level constants alongside the others; dropped the dead `_ICAL_WITH_LOCATION` test variable.

**Bot `_NULL_FOR_UPDATE`** — dismissed as a false positive (it's imported and used in `update_event`), as you noted.

I left the RECURRENCE-ID path at `ical_to_jscal.py` out of scope here since it's governed by the §4.2 override-key handling and isn't a UTC-on-TZID case; happy to revisit if you think it needs the same treatment.

⚠️ This comment is AI-generated (Claude Opus 4.8 via Claude Code) on behalf of tobixen

SashankBhamidi:
Hi, sorry for the delay. I’ve been catching up on a backlog of GitHub notifications. I’ll get to reviewing this tonight.

SashankBhamidi (approved):
Good to merge.

Thank you!

caldav/jmap/convert/ical_to_jscal.py
```diff
@@ -12,12 +12,40 @@
...
+    expressed in the recurring event's own time zone.  A UTC UNTIL/EXDATE on a
+    TZID event must therefore be converted to that zone before being formatted
+    as a (suffix-less) LocalDateTime, otherwise the series ends at the wrong
+    wall-clock time and the round-trip emits a Z-less UNTIL that RFC 5545
+    §3.3.10 forbids for TZID events.  Floating / all-day / naive values (and
+    unknown time zones) pass through unchanged.
+    """
+    if time_zone and isinstance(dt, datetime) and dt.tzinfo is not None:
+        try:
+            return dt.astimezone(ZoneInfo(time_zone)).replace(tzinfo=None)
```
niccokunzmann, line 43: @tobixen icalendar has a few more fixes for timezone names:

https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.timezone.tzp.html#icalendar.timezone.tzp.TZP.localize

If you like, let me review your code from time to time, so I can point out what to use from icalendar and other packages I maintain.

tobixen:
I hate code duplication, and I'm all for consolidating things wherever it may be consolidated.  However, the jmap code is going out through the door now and into the jmap library, so it should probably be followed up there.  That said, I probably have quite a lot of other similar timezone handling in the caldav part of the library.

PR #711, opened 2026-09-15
feat(jmap)!: re-export caldav.jmap from the standalone calendaring-jmap package

### LINKED ISSUE #10: Re-export caldav.jmap through calendaring-jmap in python-caldav
Once 1.0 is on PyPI, python-caldav needs updating so nothing breaks for existing users.

pyproject.toml gets:
```toml
[project.optional-dependencies]
jmap = ["calendaring-jmap>=1.0"]
```

`caldav/jmap/__init__.py` imports from `calendaring_jmap` and re-exports with a `DeprecationWarning`. Both import paths must keep working:

```python
from caldav.jmap import JMAPClient        # still works, deprecated
from calendaring_jmap import JMAPClient   # new canonical form
```

Do this after 1.0 is published, not before.

SashankBhamidi:
## Summary

- `caldav/jmap/` no longer carries its own copy of the JMAP client; it now re-exports the standalone [`calendaring-jmap`](https://pypi.org/project/calendaring-jmap/) package (v1.1.0), added as an optional dependency (`caldav[jmap]`).
- `from caldav.jmap import JMAPClient` still works, but now emits a `DeprecationWarning` pointing at the new canonical `from calendaring_jmap import JMAPClient`.
- `get_jmap_client()`/`get_async_jmap_client()` keep resolving configuration through `caldav.config.get_connection_params()` exactly as before.
- JMAP errors stay catchable as `DAVError`, via a conditional import on calendaring-jmap's side (pycalendar/calendaring-jmap#43, released as 1.1.0).
- Docs trimmed to a short pointer at calendaring-jmap's own ReadTheDocs.

Closes https://github.com/pycalendar/calendaring-jmap/issues/10


tobixen:
This is on my list, and I expect to have it merged to main tomorrow.  Before approving this pull request, I should ensure all tests passes.  I just released version 3.3.1, so at least the conflict in CHANGELOG needs some massaging.

Background: The JMAP logic was requested in the CalDAV-library based on an old user request, but it doesn't really belong in the CalDAV-library.  Sashank has now been reimplementing the logic in a separate JMAP-library, and the code should be yanked out of the CalDAV-libary.  To try not to break backward-compatibility, the CalDAV-library will come with an optional dependency on the JMAP-library and wrapper functions.  If it's hard to maintain backward-compatibility and one has a good reason for it, then breaking changes are OK - because the JMAP-support in CalDAV 3.x was marked up as experimental in the documentation.

tobixen:
With jmap as an optional dependency, the import will break with a bare `ModuleNotFoundError`.  I think we should wrap it around a try/except-block with descriptive error messages (`install caldav[jmap] or jmap` or something like that).

SashankBhamidi:
Wrapped the calendaring_jmap import in try/except with a descriptive error pointing at `caldav[jmap]`.

tobixen:
I asked Claude to have a look, and it did record some "breaking changes":

* **Breaking:** `caldav.jmap` no longer works on a plain `pip install caldav`.  The implementation moved to the standalone [calendaring-jmap](https://pypi.org/project/calendaring-jmap/) package, which is an optional dependency - install `caldav[jmap]` (or `calendaring-jmap`).  Importing `caldav.jmap` without it raises an `ImportError` saying so.
* **Breaking:** `caldav[jmap]` brings dependencies caldav itself does not have.  calendaring-jmap 1.1.0 requires `icalendar>=7.3.0` (caldav asks only for `icalendar>6.0.0`, so the extra raises the floor), and it requires both `niquests` and `requests` outright - so `requests` is installed even in the environments that deliberately avoid it, see [HTTP Library Configuration](https://caldav.readthedocs.io/stable/http-libraries.html).
* **Breaking:** the JMAP code is licensed differently from the rest of caldav.  caldav is `GPL-3.0-or-later OR Apache-2.0`; calendaring-jmap is `AGPL-3.0-or-later`.  The import path is unchanged, so this is easy to miss: if you relied on the Apache-2.0 option, note that the JMAP code you get through `caldav[jmap]` carries the AGPL network-copyleft obligation.  Nothing changes for users of caldav without the `jmap` extra.

The first one is already an accepted compromise, the license thing I've also explicitly accepted, but the middle one seems a bit silly.  Is there anything in the current jmap code that needs icalendar 7.3.0?  If not, could you release one version with stated support for icalendar 6?  It also seems silly to have both requests and niquests as hard dependencies in the project.  (see https://github.com/python-caldav/caldav/blob/master/docs/source/http-libraries.rst for some information on requests vs niquests)

tobixen:
Rather than bouncing a ball, I decided to ask claude to amend the pull request.  It added some few commits - and apparently broke the tests while being at it.  I've looked a bit through, some of the changes it did seems sound.  I'm not sure I'm sharing the concern that it's a crime to let test code depend on an AGPL-library.  At least things should be squashed before merging to the main branch.

SashankBhamidi:
@tobixen 7.3.0 isn't arbitrary. [GHSA-qjcq-q7h7-r74v](https://github.com/collective/icalendar/security/advisories/GHSA-qjcq-q7h7-r74v), the `VALARM REPEAT` memory/CPU exhaustion bug, is mine, I found and fixed it upstream, it's in 7.2.1. calendaring-jmap parses calendar data straight off a JMAP server, untrusted input by definition, so I'm not shipping anything below that fix.

The other reasonn is the globally-unique-TZID resolution ([issue #313](https://github.com/collective/icalendar/issues/313)), the `/freeassociation.sourceforge.net/Europe/Berlin` style TZIDs libical and Evolution emit. calendaring-jmap's own TZID normalization depends on it directly, with a test that exercises exactly that case. That landed in 7.3.0, not 7.2.x, so that's where the floor has to sit.

6.x is off the table. Because it's missing both of those, and I'm not going to loosen a floor I set for a reason just to make the pin number smaller.  Checked the [changelog](https://github.com/collective/icalendar/blob/master/CHANGES.rst) against what's coming up on my roadmap too, nothing past 7.3.0 is needed, so this holds for a while.

The `niquests`+`requests` thing, yeah, that one's just sloppy. `_http.py` is written to fall back from `niquests` to `requests`, only one should ever be a hard dependency, `pyproject.toml` just doesn't match that. I'll get it fixed on my side, not urgent enough to cut a release over right now though, mid-milestone and I'd rather the real fix land as part of a proper v2 than get bolted on as a patch to this PR.

I will let you know when I make the next release with this fix.

SashankBhamidi:
That fix is merged now, main is at [514b84c](https://github.com/pycalendar/calendaring-jmap/commit/514b84c), if you want to pin it before I cut a release:

calendaring-jmap @ git+https://github.com/pycalendar/calendaring-jmap.git@514b84c

Once I tag a release you can swap that for a normal version pin.

Issue #713, opened 2026-09-17
Credentials are never sent to servers that omit WWW-Authenticate on 401 (e.g. Yahoo Calendar)


leftover-salmon:
## Summary

When a server responds `401` **without** a `WWW-Authenticate` header, `DAVClient` never
builds an auth object and never transmits the supplied credentials. The bare 401
propagates to the caller as `AuthorizationError`, which is indistinguishable from a
genuinely rejected password — but the password was never sent.

Yahoo Calendar's CalDAV server (`caldav.calendar.yahoo.com`) behaves this way. RFC 7235
§3.1 requires `WWW-Authenticate` on a 401, so this is a server bug, but it is a server
bug that currently makes Yahoo unreachable through this library even with valid
credentials.

## Affected versions

Reproduced on **2.1.0** and **2.2.6**.

## Reproduction — no account or credentials required

The deliberately wrong password is the point: it demonstrates that `self.auth` is still
`None` *after* the request, meaning nothing was ever transmitted.

```python
import caldav

client = caldav.DAVClient(
    url="https://caldav.calendar.yahoo.com/",
    username="anyone@yahoo.com",
    password="does-not-matter",
)
try:
    client.principal()
except Exception as exc:
    print(type(exc).__name__)        # AuthorizationError

print(client.auth)                   # None  <- credential never sent
```

On 2.1.0 this additionally logs a `CRITICAL` with an `lxml.etree.XMLSyntaxError`
traceback, because Yahoo's error body is JSON where XML is expected:

```
CRITICAL:root:Expected some valid XML from the server, but got this:
b'{"error":{"code": 4003, "message":"Invalid User credential"}}'
```

The exception is caught and `AuthorizationError` is raised as above, so this is noise
rather than a second failure — but it is alarming in downstream logs and would disappear
along with the root cause. 2.2.6 does not log it.

With `auth_type="basic"` the auth object is built up front and the request succeeds
(verified against a real Yahoo account: `207`, principal `/principals/users/<user>/`,
and all calendars enumerate normally).

## Server behaviour

```
$ curl -s -D - -o /dev/null -X PROPFIND https://caldav.calendar.yahoo.com/ -H 'Depth: 0'
HTTP/2 401
content-type: application/json
server: ATS
```

No `WWW-Authenticate` header in the response. The server is otherwise a functioning
CalDAV endpoint:

```
$ curl -s -D - -o /dev/null -X OPTIONS https://caldav.calendar.yahoo.com/
[...]

PR #714, opened 2026-09-18
docs(config): document auth_type and other connection keys

tobixen:
This is a **draft** PR as for now.  Claude made something up, no QA done, no human reviews, and (while I may be wrong) I believe documentation written (or rewritten) by me is better than the AI-generated documentation.

Commit message:

```
docs(config): document auth_type and other keys

The config file page listed only six caldav_ keys and three CALDAV_
environment variables, so auth_type looked unsupported there even though
it has been a connection parameter since 3.0.  Both tables now carry it,
and a note after each says that any other CONNKEYS parameter works the
same way - with the two caveats a reader would otherwise hit: auth takes
an object and so cannot come from a file, and environment variables are
never type-converted, so CALDAV_REQUIRE_TLS=false sets the parameter to
the true string "false".

Prompt: is auth_type ('basic' vs 'digest' vs 'bearer') configurable today?  As a parameter into get_davclient and/or into the configuration file?
Followup-Prompt: What version did this go in?  Please update the doc [= the config file format documentation] in a separate work-branch and create a draft PR. (...)

Assisted-By: Claude Opus 5 <noreply@anthropic.com>
```


tobixen:
This is a partial response to #713

